PR #8, opened 2021-06-01
fix(init): python and java init is broken due to invalid versions

eladb:
Now that we have decoupled the various modules in the CDK8s, we need to use a module-specific version instead of a uniform version.

Added an integration test which runs `cdk8s init TEMPLATE` for each template.

### DIFF
--- a/src/cli/cmds/init.ts
+++ b/src/cli/cmds/init.ts
@@ -20,8 +20,6 @@ class Command implements yargs.CommandModule {
   public readonly builder = (args: yargs.Argv) => args
     .positional('TYPE', { demandOption: true, desc: 'Project type' })
     .showHelpOnFail(false)
-    .option('dist', { type: 'string', desc: 'Install dependencies from a "dist" directory (for development)' })
-    .option('cdk8s-version', { type: 'string', desc: 'The cdk8s version to use when creating the new project', default: pkg.version })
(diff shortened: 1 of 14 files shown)


src/cli/cmds/init.ts
```diff
@@ -146,6 +116,16 @@ class ModuleVersion {
...
     return this.mavenVersion;
   }
+
+  private resolveVersion(module: string): string {
+    if (module === 'cdk8s-cli') {
+      module = '../../../';
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    return require(`${module}/package.json`).version;
```
iliapolo, line 126: I'm not sure I follow this. Why would `cdk8s-core` and `cdk8s-plus-17` be installed at this point?

eladb: They are needed by the init templates so the way the CLI decides which version to take is by taking a dependency on them (previously we already had a dep on `cdk8s` and now I've added `cdk8s-plus-17`). I think it's a good way to ensure that the CLI works with these versions because now we have tests in the CLI that verify that init templates work.

Naturally, after a project is created, users can upgrade to a newer version if they want.

iliapolo: Gotcha. Sounds good.

iliapolo (approved):

PR #43, opened 2021-07-21
fix(import): cannot import props that have non-standard capitalization

eladb:
json2jsii now emits `toJson()` functions for all generated
structs which can be used to convert data back to the original
schema.

This means that we can now import APIs and CRDs even if their
property names are not properly camel cased, use underscores
or hyphens, etc.

Improve test harness to also take a snapshot of the generated typescript
output


test/import/util.ts
```diff
@@ -15,21 +18,41 @@ export function testImportMatchSnapshot(name: string, fn: () => ImportBase, opti
...
-    });
-
-    const manifest = JSON.parse((await fs.readFile(jsiiPath)).toString('utf-8'));
-
-    // patch cdk8s version in manifest because it's not stable
-    manifest.dependencies.cdk8s = '999.999.999';
-    manifest.fingerprint = '<fingerprint>';
 
-    expect(manifest).toMatchSnapshot();
+    const languages = [Language.TYPESCRIPT];
```
eladb, line 22: For now, just typescript - otherwise it takes forever and forever

iliapolo (approved):

Chriscbr:
Build error was just a transient dependency issue (I noticed the exact same build error preventing a upgrade-deps PR from being merged yesterday).

eladb:
> Build error was just a transient dependency issue (I noticed the exact same build error preventing a upgrade-deps PR from being merged yesterday).

Thanks for following up.

PR #44, opened 2021-07-23
fix(import): CRDs not including metadata field

Chriscbr:
Fixes https://github.com/cdk8s-team/cdk8s-core/issues/25.

CRDs are not strictly required to include the "metadata" field in their schemas because this field is automatically available on all kubernetes resources. Currently, if a schema didn't include it, then the field would be missing on the interfaces generated by `cdk8s import`.

This change automatically adds the metadata field to all custom resource definitions, and gives them the type `cdk8s.ApiObjectMetadata`. k8s imports will continue to retain their imports typed as `k8s.ObjectMeta`.


src/import/codegen.ts
```diff
@@ -81,6 +92,11 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
       copy.required = copy.required || [];
       copy.required = copy.required.filter(x => x !== 'apiVersion' && x !== 'kind' && x !== 'status');
 
+      if (def.custom) {
+        // add "metadata" field for all CRDs, overriding any existing typings
+        copy.properties.metadata = { $ref: '#/definitions/ApiObjectMetadata' };
```
eladb, line 97: This feels a bit blunt.. Isn't it possible for a CRD to define some extra fields in their `metadata` section? Shouldn't we merge?

Chriscbr: From my research, defining extra fields is not possible, no. The kubernetes docs says

> if metadata is specified, then only restrictions on metadata.name and metadata.generateName are allowed.

See also this related PR: https://github.com/kubernetes/kubernetes/pull/77653

It's possible that a CRD further constraints one of these existing fields (e.g. imposing that metadata.name must start with the prefix "foo"), but I think this is exceedingly rare and the fields are still behaviorally the same. The only thing that we might lose are the descriptions of existing metadata fields that some CRD owners include in their specs - but a lot of these are redundant as is.

eladb: Thanks for the details. Worth adding a note in the code that this is the behavior

src/import/codegen.ts
```diff
@@ -30,10 +30,17 @@ export interface ApiObjectDefinition {
 
 /**
  * Emits the header for a generated imports file.
+ *
+ * @param custom - whether the header is being emitted for a custom resource
+ * (imported from a CRD) or a core API object
  */
-export function emitHeader(code: CodeMaker) {
+export function emitHeader(code: CodeMaker, custom: boolean) {
```
eladb, line 37:
```suggestion
export function emitHeader(code: CodeMaker, crd: boolean) {
```

eladb (approved):
Added do not merge label

PR #46, opened 2021-07-27
fix kubectl apply -f example in template

jkataja:
- `kubectl apply -f` takes either a file or a path as argument
 - wildcard argument `dist/*.k8s.yaml` fails with multiple files in `dist/`
 - replaced wildcard argument with path

Fixes #


Chriscbr (approved):
Thanks for the fix! 😄

PR #48, opened 2021-07-29
feat: go project template and imports

### LINKED ISSUE #26: Golang support in cdk8s import
Users need to be able to generate L1 API's in golang using the CLI.

```console
cdk8s import -l go k8s
```

### LINKED ISSUE #27: Golang init template
Users should be able to:

```console
cdk8s init go-app
```

Template should contain a synthesizable project with a simple resource and test.

Chriscbr:
Resolves #26 
Resolves #27

`cdk8s init go-app` now initializes a Go app with k8s imports. When run in a directory named "my-app", the following files will be generated:

```
├── cdk8s.yaml
├── dist
│   └── my-app.k8s.yaml
├── go.mod
├── go.sum
├── help
├── imports
│   └── k8s
│       ├── internal
│       │   └── types.go
│       ├── jsii
│       │   ├── jsii.go
│       │   └── k8s-0.0.0.tgz
│       ├── k8s.go
│       ├── k8s.init.go
│       └── version
└── main.go
```

`cdk8s import <spec> --language go` now imports the k8s spec or custom resource definitions in Go. Each CRD generates a package in an "imports" directory, and package names are generated based on Go idiomatic naming conventions. 

Tested on golang v1.16.

### DIFF
--- a/src/import/base.ts
+++ b/src/import/base.ts
@@ -129,16 +129,17 @@ export abstract class ImportBase {
 
           // go!
           if (options.targetLanguage === Language.GO) {
-            const userModuleName = this.getGoModuleName(outdir);
+            const { userModuleName, userModulePath } = this.getGoModuleName(outdir);
+            const relativeDir = path.relative(userModulePath, outdir);
 
             // go package names may only consist of letters or digits.
             // underscores are allowed too, but they are less idiomatic
-            // this converts e.g. "cert-manager.path.to.url" to "certmanager"
-            const importModuleName = module.name.split('.')[0].replace(/[^A-Za-z0-9]/g, '').toLocaleLowerCase();
+            // this converts e.g. "cert-manager.path.to.url" to "certmanagerpathtourl"
+            const importModuleName = module.name.replace(/[^A-Za-z0-9]/g, '').toLocaleLowerCase();
 
             opts.golang = {
               outdir: outdir,
-              moduleName: `${userModuleName}/imports`,
+              moduleName: `${userModuleName}/${relativeDir}`,
               packageName: moduleNamePrefix ? moduleNamePrefix + '_' + importModuleName : importModuleName,
             };
           }
@@ -149,8 +150,12 @@ export abstract class ImportBase {
     }
   }
 
-  private getGoModuleName(outdir: string) {
-    outdir = path.normalize(outdir);
+  /**
+   * Traverses up directories until it finds a directory with a go.mod file,
+   * and parses the module name from the file.
+   */
+  private getGoModuleName(origOutdir: string) {
+    let outdir = path.resolve(origOutdir);
 
     while (outdir !== path.dirname(outdir)) {
       const file = path.join(outdir, 'go.mod');
@@ -163,12 +168,15 @@ export abstract class ImportBase {
           throw new Error('Invalid go.mod file - could not find module path.');
         }
 
-        return matches[1];
+        return {
+          userModuleName: matches[1],
+          userModulePath: outdir,
+        };
       }
 
       outdir = path.dirname(outdir);
     }
 
-    throw new Error(`Cannot find go.mod file within ${outdir}.`);
+    throw new Error(`Cannot find go.mod file within ${origOutdir} or any of its parent directories.`);
   }
 }
--- a/src/import/base.ts
+++ b/src/import/base.ts
@@ -137,7 +137,7 @@ export abstract class ImportBase {
             const importModuleName = module.name.split('.')[0].replace(/[^A-Za-z0-9]/g, '').toLocaleLowerCase();
 
             opts.golang = {
-              outdir: 'imports',
+              outdir: outdir,
               moduleName: `${userModuleName}/imports`,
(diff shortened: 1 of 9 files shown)


test/import/util.ts
```diff
@@ -24,6 +24,10 @@ export async function expectImportMatchSnapshot(fn: () => ImportBase, options?:
     for (const lang of languages) {
       const jsiiPath = lang === Language.TYPESCRIPT ? path.join(workdir, '.jsii') : undefined;
 
+      if (lang === Language.GO) {
+        fs.writeFileSync(path.join(workdir, 'go.mod'), 'module integtest');
+      }
```
Chriscbr, line 29: I added this in case we do want to add go snapshot tests at some point. Right now though, snapshot takes very long and if I add Go to the language list on line 22, then it takes 8+ minutes to run a full build+test on my machine, so I wasn't sure if it's worth the cost. Just say the word and I can add them though 🙂

Chriscbr:
Hmm - not necessarily blocking, but one thing I notice is that in the current implementation, setting up Go projects might be slightly awkward because it doesn't ask you for a project name (technically called the "module path"), instead providing a default value of `example.com/<directory name>`.

But since Go's import system uses full path names everywhere, if you ever wanted to rename your module (e.g. to `github.com/username/my-app`), then you'd have to update the import paths everywhere or set up a Go "replace" directive in the `go.mod` file.

Anyway, this is all to say we want feedback from Go users on this experience and/or might want to consider adding an option to `cdk8s init` to let users pass in their initial project name.

templates/go-app/cdk8s.yaml
```diff
@@ -0,0 +1,4 @@
+language: python
```
iliapolo, line 1: Copy/Paste error? :)

test/import/util.ts
iliapolo: I'm ok with skipping snapshots for go for now - feels like we should add all languages at some point and maybe parallelize in the workflow somehow.

iliapolo (changes requested):

src/import/base.ts
```diff
@@ -126,9 +127,48 @@ export abstract class ImportBase {
...
+          throw new Error('Invalid go.mod file - could not find module path.');
+        }
+
+        return matches[1];
+      }
+
+      outdir = path.dirname(outdir);
+    }
+
+    throw new Error(`Cannot find go.mod file within ${outdir}.`);
```
iliapolo, line 172: You should use the original `outdir` here otherwise it will always print `/` in the error. 

```suggestion
    throw new Error(`Cannot find go.mod file within ${origOutdir} or any of its parent directories.`);
```

src/import/base.ts
```diff
@@ -126,9 +127,48 @@ export abstract class ImportBase {
             };
           }
 
+          // go!
+          if (options.targetLanguage === Language.GO) {
+            const userModuleName = this.getGoModuleName(outdir);
+
+            // go package names may only consist of letters or digits.
+            // underscores are allowed too, but they are less idiomatic
+            // this converts e.g. "cert-manager.path.to.url" to "certmanager"
```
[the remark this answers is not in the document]
Chriscbr: Fair concern, I'll change this.

src/import/base.ts
Chriscbr: Good callout!

src/import/base.ts
```diff
@@ -126,9 +127,48 @@ export abstract class ImportBase {
...
+            const importModuleName = module.name.split('.')[0].replace(/[^A-Za-z0-9]/g, '').toLocaleLowerCase();
+
+            opts.golang = {
+              outdir: outdir,
+              moduleName: `${userModuleName}/imports`,
```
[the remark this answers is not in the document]
Chriscbr: If I omit `/imports`, then the imports are generated almost identically, except that the *import statements* become invalid. That is, we find 

```go
// imports/k8s/k8s.go

package k8s

import (
	"time"

	_init_ "example.com/hello-world/k8s/jsii" // error! this should be example.com/hello-world/imports/k8s/jsii
	_jsii_ "github.com/aws/jsii-runtime-go/runtime"

	"example.com/hello-world/k8s/internal" // error! this should be example.com/hello-world/imports/k8s/internal
	"github.com/aws/constructs-go/constructs/v3"
	"github.com/cdk8s-team/cdk8s-core-go/cdk8s"
)
```

I'm not totally clear on this, but maybe the documentation just needs to be updated. It seems like in practice, the golang transpiler expects me to provide the full module path (up to but excluding the package name, like "k8s") in this `moduleName` field. By doing this, all of the generated imports are correct.

I've updated my implementation to handle this better, and I've tested that the `cdk8s import` for go works with a wide variety of `--output` options (both passing in relative and absolute paths).

iliapolo (approved):

src/import/base.ts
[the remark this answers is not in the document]
iliapolo: Looks like there's something off with either `jsii-srmak` argument names or the implementation itself. The current solution does seem to hack around this quirk, so we'll go ahead with it and look at srcmak later on

PR #54, opened 2021-08-12
fix: missing API versions in k8s imports

### LINKED ISSUE #45: cdk import doesn't include API level for `Spec`s schema
### Description of the bug:
imported `Spec`s also should have API levels.

For example, `HorizontalPodAutoscalerSpec` have been changed from v1 ([v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#horizontalpodautoscalerspec-v1-autoscaling), [v2beta2](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#horizontalpodautoscalerspec-v2beta2-autoscaling))

I understand that V1 props doesn't have API level postfix. The generated class seems to be right.
```ts
export interface KubeHorizontalPodAutoscalerProps {
  readonly metadata?: ObjectMeta;
  readonly spec?: HorizontalPodAutoscalerSpec;
}
```

I expected that the spec should be `HorizontalPodAutoscalerSpecV2Beta2` for v2beta2, but cdk import generated this:
```ts
export interface KubeHorizontalPodAutoscalerV2Beta2Props {
  readonly metadata?: ObjectMeta;
  readonly spec?: HorizontalPodAutoscalerSpec;
}
```

Actually, `HorizontalPodAutoscalerSpec`'s comment says it is generated from v2beta2, not v1.
```ts
/**
 * HorizontalPodAutoscalerSpec describes the desired functionality of the HorizontalPodAutoscaler.
 *
 * @schema io.k8s.api.autoscaling.v2beta2.HorizontalPodAutoscalerSpec
 */
export interface HorizontalPodAutoscalerSpec {
```

There should be `HorizontalPodAutoscalerSpec`, `HorizontalPodAutoscalerSpecV2Beta1`, `HorizontalPodAutoscalerSpecV2Beta2`.

### Reproduction Steps:

```shell
cdk import k8s
```

### Error Log:
No error logs

### Environment:
  - **Framework Version**: v1.0.0-beta.3
  - **OS**: Windows 10, WSL 1, Ubuntu 20.10

### Other:

--- 

This is :bug: Bug Report

Chriscbr:
Fixes #45 

The high-level strategy behind this fix is described here: https://github.com/cdk8s-team/cdk8s-cli/issues/45#issuecomment-887753986

I tested this by running `cdk8s import` in a separate project. The main difference in the generated `k8s.ts` file is that multiple versioned interfaces are generated for non-top-level definitions that have identical names but different resources. By non-top-level definitions, we mean the generated types that don't subclass from `ApiObject`, such as `XxxSpec`.

So now as an example, `KubeDeploymentV1Beta1Props` refers to `DeploymentSpecV1Beta1` instead of `DeploymentSpec`, and `DeploymentSpecV1Beta1` is generated in addition to `DeploymentSpec` (whereas it was previously omitted). If a spec is versioned at v1, then the version postfix is implicitly omitted.

The only minor regression caused by this fix is that the "schema" values associated with these types are all overwritten, as visible in the diff:

```diff
       "datatype": true,
       "docs": Object {
         "custom": Object {
-          "schema": "io.k8s.api.rbac.v1beta1.AggregationRule",
+          "schema": "AggregationRule",
         },
         "summary": "AggregationRule describes how to locate ClusterRoles to aggregate into the ClusterRole.",
       },
```

### DIFF
--- a/src/import/k8s.ts
+++ b/src/import/k8s.ts
@@ -82,7 +82,7 @@ export class ImportKubernetesApi extends ImportBase {
       } else {
         // rename struct types from their original names to ensure that
         // differently-versioned but identically-named resources
-        // are generated as distinct interfaces (differentiated by
+        // are generated as distinct types (differentiated by
         // their version).
         const { fullVersion } = parseApiTypeName(fqn);
         const canonicalName = getCanonicalName(fqn, fullVersion === '' ? 'v1' : fullVersion);
--- a/src/import/k8s.ts
+++ b/src/import/k8s.ts
@@ -8,7 +8,7 @@ import { TypeGenerator } from 'json2jsii';
 import { ImportSpec } from '../config';
 import { download } from '../util';
 import { GenerateOptions, ImportBase } from './base';
-import { ApiObjectDefinition, emitHeader, generateConstruct, getPropsTypeName } from './codegen';
+import { ApiObjectDefinition, emitHeader, generateConstruct, getCanonicalName, getPropsTypeName } from './codegen';
 import { parseApiTypeName } from './k8s-util';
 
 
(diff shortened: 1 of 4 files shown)


src/import/k8s.ts
```diff
@@ -58,26 +58,40 @@ export class ImportKubernetesApi extends ImportBase {
...
+        // order to avoid confusion between constructs (`KubeDeployment`) and those
+        // types.
+        typeGenerator.addAlias(fqn, getPropsTypeName(obj));
+
+        // emit construct type (recursive)
+        generateConstruct(typeGenerator, obj);
+      } else {
+        // rename struct types from their original names to ensure that
+        // differently-versioned but identically-named resources
+        // are generated as distinct interfaces (differentiated by
```
Chriscbr, line 85:
```suggestion
        // are generated as distinct types (differentiated by
```

iliapolo:
@Chriscbr I took a deeper look here as we discussed and I believe there is a simpler, more appropriate fix we should push to `json2jsii`. 

Basically the offending code is here: https://github.com/cdklabs/json2jsii/blob/main/src/type-generator.ts#L500. 

It causes all versioned structs to collapse to a single one (the first on the list). I think the fix would be to apply the same logic from [here](https://github.com/cdklabs/json2jsii/blob/main/src/type-generator.ts#L485) when dealing with namespaced references, so I am preparing a fix. Let me know if you think differently.

Chriscbr:
@iliapolo Got it, I think that approach makes sense. 👍

iliapolo:
Closing in favor of https://github.com/cdklabs/json2jsii/pull/509

PR #55, opened 2021-08-17
feat: fix imports not supporting Lazy values

Chriscbr:
Related to https://github.com/cdk8s-team/cdk8s-plus-17/pull/56.


src/import/codegen.ts
```diff
@@ -159,7 +163,21 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
       code.openBlock(`public static ${MANIFEST_STATIC_METHOD}(props: ${propsTypeName}${defaultProps}): any`);
       code.open('return {');
       code.line(`...${constructName}.${GVK_STATIC},`);
-      code.line(`...toJson_${propsTypeName}(props),`);
+      code.line('...props,');
```
iliapolo, line 166: The `manifest` function should still invoke `toJson_xxx` since its a public API expected to return a valid manifest. 
We just need to change the constructor to not call this function - but instead use:

```ts
public constructor(scope: Construct, id: string, props: FooCronTabProps = {}) {

    // do not call `manifest()` here
    super(scope, id, {
      ...FooCronTab.GVK,
      ...props,
    });
  }
```

iliapolo (changes requested):
One change

MrArnoldPalmer (approved):

PR #66, opened 2021-09-07
chore: deprecate node v10

Chriscbr:
Node v10 is now deprecated and is no longer being maintained. This should unblock the failing upgrade-dependencies workflow (https://github.com/cdk8s-team/cdk8s-cli/actions/runs/1207548727).

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -11,7 +11,7 @@ export async function shell(program: string, args: string[] = [], options: Spawn
   return new Promise((ok, ko) => {
     const child = spawn(program, args, { stdio: ['inherit', 'pipe', 'inherit'], ...options });
     const data = new Array<Buffer>();
-    child.stdout.on('data', chunk => data.push(chunk));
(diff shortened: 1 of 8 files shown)


src/util.ts
```diff
@@ -11,7 +11,7 @@ export async function shell(program: string, args: string[] = [], options: Spawn
   return new Promise((ok, ko) => {
     const child = spawn(program, args, { stdio: ['inherit', 'pipe', 'inherit'], ...options });
     const data = new Array<Buffer>();
-    child.stdout.on('data', chunk => data.push(chunk));
+    child.stdout?.on('data', chunk => data.push(chunk));
```
Chriscbr, line 14: This is due to a change in @types/node. Node.js docs says:

> If the child was spawned with stdio[1] set to anything other than 'pipe', then this will be null.

You can see above stdout is set to "pipe" so the usage of stdout is still correct.

eladb (approved):

PR #67, opened 2021-09-14
feat: input validation

iliapolo:
Fixes #


src/import/k8s.ts
```diff
@@ -163,5 +163,9 @@ const X_GROUP_VERSION_KIND = 'x-kubernetes-group-version-kind';
 async function downloadSchema(apiVersion: string) {
   const url = `https://raw.githubusercontent.com/awslabs/cdk8s/master/kubernetes-schemas/v${apiVersion}/_definitions.json`;
   const output = await download(url);
-  return JSON.parse(output) as JSONSchema4;
+  try {
+    return safeParseJsonSchema(output) as JSONSchema4;
```
Chriscbr, line 167: Why not change the return type of `safeParseJsonSchema` to JSONSchema4? Or is that not technically type safe

test/import/import-k8s-util.test.ts
```diff
@@ -51,4 +52,127 @@ test('parseApiTypeName', () => {
...
+          },
+          required: [
+            'sideEffects',
+          ],
+          type: 'object',
+        },
+      },
+    };
+    const parsed = safeParseJsonSchema(JSON.stringify(schema));
+    expect(parsed.definitions.MutatingWebhook.properties.sideEffects.somethingElse).toEqual(SafeReviver.STRIPPED_VALUE);
```
Chriscbr, line 128: Not sure if I understand this test case, why is it stripped?

test/import/import-crd.test.ts
```diff
@@ -219,3 +237,120 @@ describe('classPrefix can be used to add a prefix to all construct class names',
...
+            description: 'Its ok to have spaces here',
+          },
+        },
+      },
+    };
+
+    await withTempFixture(crd, async (fixture: string) => {
+      const definitions = await ImportCustomResourceDefinition.match({ source: fixture });
+      expect(definitions?.length).toEqual(1);
+      const definition = definitions ? definitions[0] : undefined;
```
Chriscbr, line 324:
```suggestion
      const definition = definitions[0];
```

test/import/import-crd.test.ts
```diff
@@ -219,3 +237,120 @@ describe('classPrefix can be used to add a prefix to all construct class names',
...
+            description: "*/console.log('hello')/*",
+          },
+        },
+      },
+    };
+
+    await withTempFixture(crd, async (fixture: string) => {
+      const definitions = await ImportCustomResourceDefinition.match({ source: fixture });
+      expect(definitions?.length).toEqual(1);
+      const definition = definitions ? definitions[0] : undefined;
```
Chriscbr, line 267: If we know the definitions length can we just access it directly?
```suggestion
      const definition = definitions[0];
```

src/import/crd.ts
```diff
@@ -177,8 +150,45 @@ export class ImportCustomResourceDefinition extends ImportBase {
...
+  const objects = safeParseYaml(manifest, reviver);
+
+  // since the manifest can contain non crds as well, we first
+  // collect all crds and only apply a schema validatio on them.
+
+  const crds: any[] = [];
+
+  function collectCRDs(objs: any[]) {
+    for (const obj of objs.filter(o => o)) {
+      if (obj.kind === CRD_KIND) {
```
Chriscbr, line 170: nit: Would be nice if we could make this more type safe by using `objs: unknown[]`

Chriscbr (changes requested):
A few comments, nit's are optional as usual

src/import/crd.ts
iliapolo: Not sure I see the value. What type assertion do you see missing here? We aren't using `obj.kind` because actually checking its value.

test/import/import-crd.test.ts
iliapolo: Changed these tests to actually produce the generated `.ts` tests and compare snapshots.

test/import/import-crd.test.ts
iliapolo: https://github.com/cdk8s-team/cdk8s-cli/pull/67#discussion_r709107115

test/import/import-k8s-util.test.ts
iliapolo: Added a comment.

src/import/k8s.ts
iliapolo: Should be fine. Changed 👍

src/import/crd.ts
```diff
@@ -177,8 +150,45 @@ export class ImportCustomResourceDefinition extends ImportBase {
   }
 }
 
-function assert(condition: boolean, message: string) {
-  if (!condition) {
-    throw new Error(`invalid CustomResourceDefinition manifest: ${message}`);
+export function safeParseCrds(manifest: string): any[] {
```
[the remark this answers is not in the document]
iliapolo: Yeap, changed.

scripts/crd.schema.ts
```diff
@@ -0,0 +1,28 @@
...
+import * as fs from 'fs';
+
+const out = path.join(__dirname, '..', 'src', 'schemas', `crd.schema.json`);
+
+const settings: Partial<tjs.Args> = {
+  required: true,
+  ref: true,
+  topRef: true,
+  noExtraProps: false,
+  out
```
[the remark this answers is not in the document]
iliapolo: Thanks, removed a few un needed ones in the process 👍

.projenrc.js
```diff
@@ -51,6 +53,14 @@ const project = new TypeScriptProject({
...
   autoApproveUpgrades: true,
+  tsconfig: {
+    include: ['src/schemas/*.json'],
+  },
 });
 
+const schemas = project.addTask('schemas');
+schemas.exec('ts-node scripts/crd.schema.ts');
+
+project.buildTask.spawn(schemas);
```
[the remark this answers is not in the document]
iliapolo: You're right - fixed.

src/import/crd.ts
```diff
@@ -102,49 +105,29 @@ export class CustomResourceDefinition {
...
     const { source } = importSpec;
     const manifest = await download(source);
-    return yaml.parseAllDocuments(manifest).map((doc: yaml.Document) => doc.toJSON());
+    return new ImportCustomResourceDefinition(safeParseCrds(manifest));
   }
 
   private readonly groups: Record<string, CustomResourceDefinition[]> = { };
 
-  constructor(manifest: ManifestObjectDefinition[]) {
+  private constructor(manifest: ManifestObjectDefinition[]) {
```
iliapolo, line 116: Made this private so that this object cannot be created without input validation.

src/import/crd.ts
```diff
@@ -102,49 +105,29 @@ export class CustomResourceDefinition {
...
     super();
 
     const crds: Record<string, CustomResourceDefinition> = { };
     const groups: Record<string, CustomResourceDefinition[]> = { };
 
-    const extractCRDs = (objects: ManifestObjectDefinition[] = []) => {
```
iliapolo, line 119: This was moved to the yaml parsing logic since we need to perform schema validation on each CRD.

src/import/crds-dev.ts
```diff
@@ -17,26 +14,12 @@ import { ImportCustomResourceDefinition, ManifestObjectDefinition } from './crd'
 /**
  * Matches a https://doc.crds.dev repo
  *
- *  - ManifestObjectDefinition[] if found
+ *  - url if found
  *  - undefined if not
  *
  * @param source
  */
-export async function importCrdsDevRepoMatch(importSpec: ImportSpec): Promise<undefined | ManifestObjectDefinition[]> {
```
iliapolo, line 25: A consequence of changing the `ImportCustomResourceDefinition` constructor to private.

Chriscbr (approved):

PR #119, opened 2021-11-08
Allow @ in key name

jvassev:
Envoy-related api objects use `@type` a lot:
```yaml
kind: EnvoyFilter
metadata:
  name: metadata-exchange-1.8
  namespace: istio-system
spec:
  configPatches:
    - applyTo: HTTP_FILTER
      match:
        context: SIDECAR_INBOUND
        proxy:
          proxyVersion: '^1\.8.*'
        listener:
          filterChain:
            filter:
              name: "envoy.filters.network.http_connection_manager"
      patch:
        operation: INSERT_BEFORE
        value:
          name: istio.metadata_exchange
          typed_config:
            "@type": type.googleapis.com/udpa.type.v1.TypedStruct
            type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
```

Without this patch, one would get an error similar to this:
```
Error: Key '@type' contains non standard characters (Must match regex '/^(\w|\.|\/|-|#|,)*$/'
```

Fixes #


jvassev:
Any updates on this issue?

PR #123, opened 2021-11-12
feat(templates): `ts-node` by default

bmacher:
CDK switched from node to `ts-node` a while ago. So I thought cdk8s could do the same, hence this PR. I enabled `noEmit` in the tsconfig to omit the transpiled output. I think the core [cdk8s](https://github.com/cdk8s-team/cdk8s) repository needs `ts-node` as a dev dependency when I understand your code correctly (see below). I can create a PR for that too, if you want.

https://github.com/cdk8s-team/cdk8s-cli/blob/f5b40896e32968c9a0d319353963478d9e66ca72/src/cli/cmds/init.ts#L44

CDK: https://github.com/aws/aws-cdk/blob/1e88655a609ec83af5276c3af9af23472a9fad73/packages/aws-cdk/lib/init-templates/v2/app/typescript/cdk.template.json#L2


bmacher:
Hi, any thoughts on this?

bmacher:
I get your point. But I would argue that one does not `synth` every change. However, I like the idea to give those people the possibility to use the "traditional" way.

bmacher:
Anything blocking this PR?

iliapolo:
@bmacher Nothing blocking per se. We just haven't gotten around to fully considering the implications of this. But I think we should be good here, i'll give this another thought.

bmacher:
Only commenting to avoid the PR being closed.

bmacher:
Only commenting to avoid the PR being closed.

bmacher:
🚀

bmacher:
Groundhog Day 1

bmacher:
Groundhog Day 2

iliapolo:
@bmacher Apologies for punting this for so long. I'm willing to merge this. Are you able to pick this back up and resolve the conflicts?

bmacher:
@iliapolo, should now be ready to merge.

iliapolo:
@bmacher Can you rebase your branch and create the PR against `2.x`? We will auto backport it to v1.

bmacher:
@iliapolo, I set the base to `2.x` and rebased my branch. Hopefully, everything is fine now.

templates/typescript-app/tsconfig.json
```diff
@@ -10,6 +10,7 @@
       "es2016"
     ],
     "module": "CommonJS",
+    "noEmit": true,
```
iliapolo:
```suggestion
    "noEmit": false,
```

Lets keep emitting `.js` files in case the app ends up being published.

iliapolo (approved):
@bmacher Thanks for this 👍 

Notice I change `tsconfig` to emit javascript files so that the app can be easily published.

PR #145, opened 2021-12-08
chore: list Homebrew installation method in README

Chriscbr:


iliapolo:
This is...peculiar. I do see those version on brew, but we are not explicitly publishing them. 
I'd like to find an explanation to this before we publicize it.

Chriscbr:
@iliapolo  I believe I found the source - it looks like the Homebrew maintainers have an "autobump" GitHub workflow that includes cdk8s in it. This commit in specific added cdk8s to the list of autobump targets: https://github.com/Homebrew/homebrew-core/commit/ed312b910e8a1620bbfefa50bee2c2635d53157f.

iliapolo (approved):

PR #176, opened 2022-01-14
fix: List files generated in sub-folders

camba1:
PR addresses the following issue:

When cdk8s synth generates files in sub-folders of the output folder, the cli thinks that no files where generated and prints 'No manifests synthesized' in the console. This is because the CLI is only looking for files in the output directory itself and not its subdirectories.

The PR changes the CLI to search for files both in the output folder and its sub-folders

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,4 +1,5 @@
 import { spawn, SpawnOptions } from 'child_process';
+import { promises } from 'fs';
 import * as http from 'http';
 import * as https from 'https';
 import * as os from 'os';
@@ -50,11 +51,12 @@ export async function synthApp(command: string, outdir: string) {
   }
 
   let found = false;
-  for (const file of await fs.readdir(outdir)) {
-    if (file.endsWith('.k8s.yaml')) {
-      console.log(`${outdir}/${file}`);
-      found = true;
+  const yamlFiles = await getFiles(outdir);
+  if (yamlFiles?.length) {
+    for (const yamlFile of yamlFiles) {
+      console.log(yamlFile);
     }
+    found = true;
   }
 
   if (!found) {
@@ -128,4 +130,25 @@ export async function download(url: string): Promise<string> {
     req.once('error', ko);
     req.end();
   });
+}
+
+async function getFiles(filePath :string) {
+  const entries = await promises.readdir(filePath, { withFileTypes: true });
+
+  // Get files within the current directory
+  const files = entries
+    .filter(file => (!file.isDirectory() && file.name.endsWith('.k8s.yaml')))
+    .map(file => (filePath + '/' + file.name));
+
+  // Get sub-folders within the current folder
+  const folders = entries.filter(folder => folder.isDirectory());
+
+  for (const folder of folders) {
+    /*
+        Get files in sub-folders
+      */
+    files.push(...await getFiles(`${filePath}/${folder.name}`));
+  }
+
+  return files;
 }


src/util.ts
```diff
@@ -129,3 +131,24 @@ export async function download(url: string): Promise<string> {
     req.end();
   });
 }
+
+async function getFiles(filePath :string) {
```
Chriscbr, line 135: Please annotate this with the appropriate return type (I think it's `string[]`?)
```suggestion
async function getFiles(filePath: string): string[] {
```

src/util.ts
```diff
@@ -129,3 +131,24 @@ export async function download(url: string): Promise<string> {
...
+    .filter(file => (!file.isDirectory() && file.name.endsWith('.k8s.yaml')))
+    .map(file => (filePath + '/' + file.name));
+
+  // Get sub-folders within the current folder
+  const folders = entries.filter(folder => folder.isDirectory());
+
+  for (const folder of folders) {
+    /*
+        Get files in sub-folders
+      */
```
Chriscbr, line 149: We can remove this comment

Chriscbr (changes requested):
Can we please add a unit test for the new `getFiles` function?

src/util.ts
camba1: I made the changes to remove the comment and annotate the function with the return type.
On adding the unit tests, sure thing. However I was looking for the tests for the existing functionality as I could probably reuse most of that but I cannot find it. Could you point me to it? I see a test/synth folder but that does not seem to be it
Thanks

Chriscbr: I'm not sure if there are any existing tests for this. Either way `getFiles` is self-contained so you can just put any tests in a new test file (like `util.test.ts`) within the tests directory.

It should be possible to use code like this to create temporary directories and files within a test case:

```ts
const basedir: string = await fs.mkdtemp(path.join(os.tmpdir(), 'something'));
const subdir: string = await fs.mkdirp(path.join(basedir, 'something-else'));
await fs.writeFile(path.join(basedir, 'example.yml'), 'contents');
```

camba1: @Chriscbr I added the test for the new function

Thanks!

test/util.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { tmpdir } from 'os';
+import path from 'path';
+import { getFiles } from '../src/util';
+
+describe('getFiles', () => {
+
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
```
Chriscbr, line 11: It doesn't look like anything is being mocked, so it's safe to remove this. 
```suggestion
```

test/util.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    await promises.writeFile(path.join(PopulatedSubDir2, `example3.${extensionFilter}`), 'contents');
+    await promises.writeFile(path.join(PopulatedSubDir2, 'example4.txt'), 'contents');
+
+
+    // THEN
+
+    // Get the contents of the different folders
+    const noFiles : string[] = await getFiles(emptySubDir);
+    const yamlFiles : string[] = await getFiles(PopulatedSubDir);
+    const NoFilesEither : string[] = await getFiles(nonDir);
```
Chriscbr, line 38: style - no space between variable name and ":"
```suggestion
    const noFiles: string[] = await getFiles(emptySubDir);
    const yamlFiles: string[] = await getFiles(PopulatedSubDir);
    const noFilesEither: string[] = await getFiles(nonDir);
```

test/util.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    //WHEN we have this folder structure
+
+    // Only look for files of this type
+    const extensionFilter = 'k8s.yaml';
+    // Create folders
+    const baseDir: string = await promises.mkdtemp(path.join(tmpdir(), 'cdk8s-getFiles-test'));
+    const emptySubDir: string = await promises.mkdtemp(path.join(baseDir, 'emptyDir'));
+    const PopulatedSubDir: string = await promises.mkdtemp(path.join(baseDir, 'populatedDir'));
+    const PopulatedSubDir2: string = await promises.mkdtemp(path.join(PopulatedSubDir, 'populatedDir2'));
```
Chriscbr, line 23: We prefer to use lowerCamelCase for variables to be consistent throughout our codebase:
```suggestion
    const populatedSubDir: string = await promises.mkdtemp(path.join(baseDir, 'populatedDir'));
    const populatedSubDir2: string = await promises.mkdtemp(path.join(populatedSubDir, 'populatedDir2'));
```

Chriscbr (changes requested):
Looking great, just a few more comments.

test/util.test.ts
camba1: Thanks for the suggestions @Chriscbr! The changes are done and committed

Chriscbr (approved):

PR #198, opened 2022-02-15
chore: reschedule upgrades to make build failures less likely

Chriscbr:
New versions of cdk8s and cdk8s-plus are published/released to different languages at slightly different times because the publishing happens on different package registries.

However, this causes our integration tests to sporadically fail because:

1. upgrade-dependencies workflows are triggered at roughly the same time every day for all repositories
2. our integration tests expect if cdk8s@x.y.z is available on NPM, then cdk8s@x.y.z is also available on PyPI, Maven, etc.

Currently, whenever this issue happens, it can be easily fixed by re-running the workflow again at any time that isn't when a new version of cdk8s or cdk8s-plus is being released. (so any time that's not close to 00:00 UTC). But this shouldn't have to be a manual effort.

This PR mitigates (1) and decreases the chance of upgrade workflows failing by running the workflow at an offset time compared to the other repos.

### DIFF
--- a/.github/workflows/upgrade-main.yml
+++ b/.github/workflows/upgrade-main.yml
@@ -4,7 +4,7 @@ name: upgrade-main
 on:
   workflow_dispatch: {}
   schedule:
-    - cron: 0 0 * * *
+    - cron: 0 1 * * *
 jobs:
   upgrade:
     name: Upgrade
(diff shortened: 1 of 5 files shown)


eladb (approved):
Is there a better way?

PR #270, opened 2022-04-26
chore: remove jsii-superchain for CI

Chriscbr:
Fix some currently failing builds caused by jsii-superchain "latest" tag no longer being available. Instead of switching to another version, let's just drop it as a requirement.

### DIFF
--- a/test/import/import-crd.test.ts
+++ b/test/import/import-crd.test.ts
@@ -9,7 +9,7 @@ import { testImportMatchSnapshot } from './util';
 const fixtures = path.join(__dirname, 'fixtures');
 
 async function withTempFixture(data: any, test: (fixture: string, cwd: string) => Promise<void>) {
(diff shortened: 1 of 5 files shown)


test/import/import-crd.test.ts
```diff
@@ -9,7 +9,7 @@ import { testImportMatchSnapshot } from './util';
 const fixtures = path.join(__dirname, 'fixtures');
 
 async function withTempFixture(data: any, test: (fixture: string, cwd: string) => Promise<void>) {
-  const tempDir = fs.mkdtempSync(os.tmpdir());
+  const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'cdk8s-import-test'));
```
Chriscbr, line 12: See https://github.com/aws/aws-cdk/pull/19989 for more info

comcalvi (approved):

PR #278, opened 2022-05-16
feat: introduce v2.x

iliapolo:
Since we have two major versions of `cdk8s`, we need the cli to support initializing templates that use either one. 

We could have done this by adding a `--cdk8s-version` flag to the cli and thats it. However:

- We want the default to be `cdk8s` v2, and if we change the default, thats a breaking change.
- Keeping `cdk8s-cli` at v1 is going to cause some confusion with users since other cdk8s versions will be v2.

Granted, introducing a new release branch creates a maintenance burden, but I feel we can handle it with automatic backports for a while, like we do with [`cdk8s-core`](https://github.com/cdk8s-team/cdk8s-core/pull/482) and `cdk8s-plus`.


iliapolo:
Note that this PR is going to fail building (due to version conflicts in `constructs`) until https://github.com/cdk8s-team/cdk8s-plus/pull/696 is merged and we can bump `cdk8s-plus` version.

.vscode/settings.json
```diff
@@ -0,0 +1,31 @@
+{
```
Chriscbr: Looks like this was added in error

Chriscbr: Maybe we should add this to the default gitignore in projen?

Chriscbr (approved):
LGTM, just one small nit

PR #324, opened 2022-06-08
fix(import): some resource don't place `metadata` immediately after `apiVersion` and `kind`

### LINKED ISSUE #323: Generated Props schemas field ordering does not match expectations from cdk8s-core
Per https://github.com/cdk8s-team/cdk8s-core/blob/58fb8c0882ddd95a9b9dedb4107e12f601443cf4/src/api-object.ts#L185, it looks like we expect that the `ApiVersion`, `Kind`, and `Metadata` fields are populated first, and placed at the top of a generated YAML document.

At one point in the past, this was the behavior, but with the introduction of per-object `toJson_...` functions, this behavior changed. Now, the ordering of the fields matches whatever is in the K8S API schemas in cdk8s/kubernetes-schemas.

My expectation is that the behavior of the generated `toJson_` functions would match the `ApiObject.toJson()` function's behavior.

Expected manifest:
```
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app: myservice
  name: my-config-map
  namespace: myservice
data:
  mydata: goes here
```

Actual manifest:
```
apiVersion: v1
kind: ConfigMap
data:
  mydata: goes here
metadata:
  labels:
    app: myservice
  name: my-config-map
  namespace: myservice
```

Especially in the case of config maps. longer data will force the metadata off screen and makes reading generated manifests more difficult and error prone.

heckler1:
Fixes #323 

Per the description of #323, this change places the `Metadata` field at the beginning of the generated interface structs, to match the behavior of the cdk8s-core `ApiObject.toJson()` function. We don't bother with `ApiVersion` or `Kind`, as those properties are deleted.


heckler1:
Bumping this, was hoping someone could take a look

heckler1:
@iliapolo is this something you could help me out with?

iliapolo:
@heckler1 Yeah i'll take a look at this next week. Thanks! 👍

Naumel:
@Mergifyio update

heckler1:
@iliapolo Hi there, just checking in on this

heckler1:
Hi there, just a quick bump to see if you had any thoughts on this?

iliapolo (approved):
@heckler1 Thanks for this! And apologies for the delay.

PR #387, opened 2022-07-12
fix(import): missing versions when importing multi-version crd

### LINKED ISSUE #386: Not importing all versions of the CRDs when multiple versions are declared and defined
### Description of the bug:

When running `cdk8s import versioned_crd.yaml`, the code generated is not available for all the versions referred in the CRD. Or in my case, the code generated was for the v1apha1 version but not for the v1beta1, where both the versions are declared and defined in that CRD file.

I was using the Capsule CRD linked [here](https://github.com/clastix/capsule/blob/master/charts/capsule/crds/tenant-crd.yaml)

### Reproduction Steps:

Step 1: Create a cdk8s project in GoLang.
Step 2: Download the CRD from [here](https://github.com/clastix/capsule/blob/master/charts/capsule/crds/tenant-crd.yaml) to local. {I downloaded locally}
Step 3: Run `cdk8s import tenant-crd.yaml`
Step 4: When you validate, the configurations are for v1alpha1 only.

  - **Framework Version**: 1.0.112
  - **OS**: MacOs

This is :bug: Bug Report

iliapolo:
For some reason, we were assuming a CRD can only contain a single version. We were extracting the first version from the `versions` property and operating solely on that.

This PR considers all versions in the CRD and appends the version to the FQN to avoid name collisions. 
Note that for backwards compatibility, the first version remains un-sufixed. 

So, for a CRD with kind `MyCrd` and versions `['v1alpha1', 'v1beta1']`, we will get:

```ts
// for the first version - unsufixed
export class MyCrd extends ApiObject { ... };
export interface MyCrdProps { 
  readonly metadata?: ApiObjectMetadata;
  readonly spec?: MyCrdSpec;
};

// for the second version - sufixed
export class MyCrdV1Beta1 extends ApiObject { ... };
export interface MyCrdV1Beta1Props {
  readonly metadata?: ApiObjectMetadata;
  readonly spec?: MyCrdV1Beta1Spec;
}
```

Where we only used to have:

```ts
export class MyCrd extends ApiObject { ... };
export interface MyCrdProps { 
  readonly metadata?: ApiObjectMetadata;
  readonly spec?: MyCrdSpec;
};
```

Fixes https://github.com/cdk8s-team/cdk8s-cli/issues/386


test/import/import-crd.test.ts
```diff
@@ -464,4 +464,17 @@ test('given a prefix, we can import two crds with the same group id', async () =
...
+});
+
+test('can import multiple versions from the same crd', async () => {
+  const fixture = path.join(fixtures, 'multi_version_tenant_crd.yaml');
+  const importer = await ImportCustomResourceDefinition.fromSpec({ source: fixture });
+  await importer.import({ targetLanguage: Language.TYPESCRIPT, outdir: '/tmp/test-tenant' });
+});
+
+
+// test('crds-dev', async () => {
```
vinayak-kukreja: I see these are commented, can these be removed?

iliapolo: Good catch. Removed 👍

vinayak-kukreja (approved):

PR #362, opened 2022-06-28
fix: typescript-app template broken for npm > 7

iliapolo:
The `typescript-app` template has a few npm scripts in its `package.json` that run `tsc`. 

By default, the `tsc` command resolves to `./node_modules/.bin/tsc`. This used to be ok, until a [change](https://github.com/aws/jsii/pull/3602) made it so this bin script gets overridden by a different typescript version than the one declared in the template. A [fix](https://github.com/aws/jsii/pull/3623) has already been merged, which should take of this issue. However, we can also fix it in our template by having the scripts point to the expected `tsc` path.

When investigating this issue, it also struck me as weird this wasn't detected in the cli init templates tests. 
This is because up until now, our init template were using the source code of the CLI to run against, instead of the tarball.
In addition, we did not have any tests executed against newer versions of node, where npm got a major version bump and underwent some substantial behavior changes.

So, this PR also adds the following:

- Treat init template tests as integration tests that are executed against the packaged tarball.
- Run all init templates tests in parallel of build on node 14 in a dedicated workflow
- Run only typescript-app template in parallel of build on node 16 and node 18 in a dedicated workflow

Also, switch to typescript for projen file.

Fixes https://github.com/cdk8s-team/cdk8s-cli/issues/363

PR #421, opened 2022-07-27
fix: typescript app template does not compile on Windows

RomainMuller:
cdk8s-team/cdk8s-cli#362 made the `compile` and `watch` scripts of the
`typescript-app` template use a relative path to the `tsc` compiler,
however such a form does not work as intended on Windows...

Instead of doing this, explicitly declare a `devDependency` on the
`typescript` compiler and return to using just `tsc`. Also, pass the
`--build` argument, which speeds up incremental builds considerably.


templates/typescript-app/package.json
```diff
@@ -8,11 +8,14 @@
...
-    "watch": "./node_modules/typescript/bin/tsc -w",
+    "compile": "tsc --build",
+    "watch": "tsc --build -w",
     "test": "jest",
     "build": "npm run compile && npm run test && npm run synth",
     "upgrade": "npm i cdk8s@latest cdk8s-cli@latest",
     "upgrade:next": "npm i cdk8s@next cdk8s-cli@next"
+  },
+  "devDependencies": {
+    "typescript": "~4.7.4"
```
iliapolo: This actually isn't need. We currently have a somewhat round about way of installing dev dependencies in the templates:

https://github.com/cdk8s-team/cdk8s-cli/blob/2e590cff9c14cad92beed9a623512852d4de5a13/templates/typescript-app/.hooks.sscaff.js#L16-L22

So we if just remove this we should be good. 

```suggestion
```

iliapolo (approved):

iliapolo:
@Mergifyio refresh

PR #479, opened 2022-08-24
feat: manifest validation plugins

### LINKED ISSUE #460: Validation plugins mechanism
As part of https://github.com/cdk8s-team/cdk8s/discussions/936, we need the CLI to support a basic plugin mechanism for third-parties to implement validation plugins:

Example potential configuration:

```yaml
validations:
  - package: cdk8s-datree-plugin
    class: DatreeValidation
    properties:
      policy: ./my-policy.json
```

This configuration can also be located in a remote URL or a local file:

```yaml
validations: /path/to/file
```

```yaml
validations: http://url/to/file
```

In this case, the contents of the file should be fetched.

The plugin mechanism should:

1. Define the interface plugins should implement.
2. Be able to dynamically load/instantiate classes and call a well known method on them.
3. Wait for the plugin to return a result, and print a report to the user.

iliapolo:
This PR implements a new plugin mechanism for post synthesis manifest validation.

Documentation:

- https://github.com/cdk8s-team/cdk8s-validation-plugin-example/pull/1
- https://github.com/cdk8s-team/cdk8s/pull/966

Resolves https://github.com/cdk8s-team/cdk8s-cli/issues/460


iliapolo:
This PR will remain in draft until we integrate with the plugin that Datree are writing, just to make sure all the wires are properly connected. Reviews are still welcome though.

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+export interface Package {
+
+  readonly module: any;
+  readonly pkg: string;
+  readonly version: string;
+  readonly path: string;
+
+}
+
+export interface PluginManagerLoadOptions {
```
vinayak-kukreja, line 61: This looks similar to `ValidationConfig` interface. Is there a reason for them to be similar?

src/plugins/validation.ts
```diff
@@ -0,0 +1,259 @@
...
+    if (!this._status) {
+      throw new Error('Unable to determine report status: Report is incomplete. Call \'report.submit\'');
+    }
+    return this._status === 'success';
+  }
+
+  /**
+   * Transform the report to a well formatted table string.
+   */
+  public toTable(): string {
```
vinayak-kukreja: Nice :)

test/synth/synth-stdout.test.ts
```diff
@@ -1,7 +1,314 @@
-// eslint-disable-next-line
-const cmd = require('../../src/cli/cmds/synth');
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import * as yaml from 'yaml';
+import { Config, ValidationConfig } from '../../src/config';
+import { mkdtemp } from '../../src/util';
+
+beforeEach(() => {
```
vinayak-kukreja, line 7: Is this to reset stdout?

iliapolo: Nope. Its to reset `require.cache` so that every test can reacquire the `../../src/cli/cmds/synth` module, as if it was the first time, just like a CLI invocation.

src/plugins/_manager.ts
iliapolo: Yes they are basically the same...It might make sense to couple them with something like:

```ts
export interface ValidationConfig extends PluginConfig
```

Let me think about it :)

.projenrc.ts
```diff
@@ -80,6 +84,14 @@ const project = new typescript.TypeScriptProject({
   },
 });
 
+//
+// see https://nodejs.org/api/packages.html#exports
+project.addFields({
+  exports: {
+    './plugins': './lib/plugins/index.js',
+  },
```
RomainMuller, line 94: If that is your only export... maybe this could have been your `entrypoint` instead? Not that it makes a big difference (this is perhaps more future-proof)

src/cli/cmds/synth.ts
```diff
@@ -14,32 +15,61 @@ class Command implements yargs.CommandModule {
   public readonly builder = (args: yargs.Argv) => args
     .option('app', { default: config.app, required: true, desc: 'Command to use in order to execute cdk8s app', alias: 'a' })
     .option('output', { default: config.output, required: false, desc: 'Output directory', alias: 'o' })
-    .option('stdout', { type: 'boolean', required: false, desc: 'Write synthesized manifests to STDOUT instead of the output directory', alias: 'p' });
+    .option('stdout', { type: 'boolean', required: false, desc: 'Write synthesized manifests to STDOUT instead of the output directory', alias: 'p' })
+    .option('plugins-dir', { default: config.pluginsDirectory, required: false, desc: 'Directory to store cdk8s plugins.' })
+    .option('validate', { type: 'boolean', default: true, required: false, desc: 'Apply validation plugins on the resulting manifests (use --no-validate to disable)' });
+  ;
```
RomainMuller:
```suggestion
    .option('validate', { type: 'boolean', default: true, required: false, desc: 'Apply validation plugins on the resulting manifests (use --no-validate to disable)' });
```

src/cli/cmds/synth.ts
```diff
@@ -14,32 +15,61 @@ class Command implements yargs.CommandModule {
...
 
-    if (outdir !== config.output && stdout) {
+    if (outdir && outdir !== config.output && stdout) {
       throw new Error('\'--output\' and \'--stdout\' are mutually exclusive. Please only use one.');
     }
 
-    await fs.remove(outdir);
+    if (outdir) {
+      await fs.remove(outdir);
+    }
```
RomainMuller: FYI - you typically don't need `fs-extras` anymore with Node 14+...

Here, for example, if you want "recursive, idempotent rm", you can:

```suggestion
    await fs.rm(outdir, { force: true, recursive: true });
```

src/config.ts
```diff
@@ -9,15 +11,27 @@ export interface ImportSpec {
...
   readonly language?: Language;
   readonly output?: string;
   readonly imports?: string[];
+  readonly pluginsDirectory?: string;
+  readonly validations?: string | ValidationConfig[];
 }
 
 const DEFAULTS: Config = {
   output: 'dist',
+  pluginsDirectory: path.join(os.homedir(), '.cdk8s', 'plugins'),
```
RomainMuller, line 34: That's very Linux-y, but going to be awkward on Windows, and (but somewhat less) macOS. I'd recommend platform-specific defaults (although feel free to ignore me here)

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Error indicating a module was found but with an unexpected version.
+ */
+export class ModuleVersionMismatchError extends Error {
+
+  public readonly code: string = MODULE_VERSION_MISMATCH_ERROR_CODE;
+
+  constructor(pkg: string, expected: string, actual: string) {
+    super(`Version mismatch for package ${pkg}. Found ${actual} but requested ${expected}`);
+  }
```
RomainMuller: This is fine, but you could as well:

```suggestion
  constructor(
    public readonly pkg: string,
    public readonly expected: string,
    public readonly actual: string,
  ) {
    super(`Version mismatch for package ${pkg}. Found ${actual} but requested ${expected}`);
  }
```

src/plugins/_manager.ts
RomainMuller, line 61: Why is there no empty lines within the declaration of this interface like the others? Any reason or just an omission?

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+}
+
+export interface Plugin {
+
+  readonly instance: any;
```
RomainMuller: There's a bunch of `any` in these interfaces here... and I'd really recommend getting out of the habit of using `any` and instead using `unknown`, which is much safer.

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+  readonly properties?: { [key: string]: any };
+}
+
+export class PluginManager {
+
+  private readonly dir: string;
+
+  constructor(dir: string) {
+    this.dir = dir;
+  }
```
RomainMuller: Why not just:

```suggestion
  constructor(private readonly dir: string) {}
```

IMHO the only reasons to do the assignment explicitly is:
1. You're doing some validations pre-assignment
2. You're using a `#privateField`, and these can't be declared in the constructor signature

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (!clazz) {
+      throw new Error(`Unable to locate class '${options.class}' in package '${options.pkg}@${options.version}'. Are you sure you exported it?`);
+    }
+
+    return { instance: new clazz(options.properties ?? {}), class: options.class, ...pkg };
+  }
+
+  private loadPackage(pkg: string, version: string, installEnv: { [key: string]: any }): Package {
+
+    if (isNaN(parseInt(version.charAt(0)))) {
```
RomainMuller: It's a bit of a smell to use `parseInt` without specifying a base, as the default base is platform-dependent and can be surprising.

```suggestion
    if (isNaN(parseInt(version.charAt(0), 10))) {
```

Also - what is `version` supposed to be like? I guess you mean for it to be the npm package version? In this case, I have to break unfortunate news to you... `npm` accepts the following version range notations:
- `^1.2.3` (you got this covered!)
- `1.2.x`
- `1.2`
- `1`
- Possibly some more...

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    if (proto) {
+      // urls are not supported because they don't provide a name with which we can 'require' the module.
+      // if needed, we can make the loader smarter and enable this, but not for now.
+      throw new Error(`Unsupported package reference: ${pkg}. Can either be an NPM package name, or a local path to a directory`);
+    }
+
+    const local = path.isAbsolute(pkg)
+      // relative paths must start with '.' because otherise they
+      // are indistinguishable from npm package names.
```
RomainMuller: Not necessarily... npm package names may only include ONE `/`, IF the segment before it starts with an `@`.

Now - I support requiring "local path" to be configured using absolute or "."-relative only, as this is more legible.

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      if (![MODULE_NOT_FOUND_ERROR_CODE, MODULE_VERSION_MISMATCH_ERROR_CODE].includes(e.code)) {
+        // some unexpected error
+        throw e;
+      }
+
+      if (local) {
+        // if a local plugin is missing, nothing we can do about it
+        throw e;
+      }
```
RomainMuller: I would have:

```suggestion
      if (local || ![MODULE_NOT_FOUND_ERROR_CODE, MODULE_VERSION_MISMATCH_ERROR_CODE].includes(e.code)) {
        // some unexpected error
        throw e;
      }
```

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+      'install', `${pkg}@${version}`,
+      '--no-save',
+      '--prefix', pluginDir,
+    ].join(' ');
+
+    const finalEnv = { ...process.env };
+    for (const [key, value] of Object.entries(env)) {
+      finalEnv[key] = typeof value === 'string' ? key : JSON.stringify(value);
+    }
+    child.execSync(command, { stdio: ['ignore', 'pipe', 'pipe'], env: finalEnv });
```
RomainMuller, line 173: Might not work properly on Windows if you don't specify `shell: true` (although not sure it applies in this case)

```suggestion
    child.execSync(command, { shell: true, stdio: ['ignore', 'pipe', 'pipe'], env: finalEnv });
```

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private require(pkg: string, version: string): Package {
+
+    const modulePath = require.resolve(pkg);
+
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const module = require(modulePath);
```
RomainMuller, line 181: At this point you've effectively executed the module... all of it's side effects have happened... But you're going to be checking if that's the right version _after_ doing this? This is awkward...

src/plugins/_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (fs.existsSync(packageJsonPath)) {
+      return packageJsonPath;
+    }
+
+    if (path.dirname(fdp) === fdp) {
+      throw new Error(`Unable to locate package.json file. Searched in: ${searched.join(os.EOL)}`);
+    }
+
+    return this.findPackageJson(path.dirname(fdp), [...searched, packageJsonPath]);
+  }
```
RomainMuller: This whole function might not work the way you expect it to in some environments. For example, when using Yarn PnP, there might not be anything on disk for you to read this way...

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  private readonly _report: ValidationReport;
+  private readonly _logger: ValidationLogger;
+
+  constructor(
+
+    /**
+     * The list of manifests to validate.
+     */
+    public readonly manifests: string[],
```
RomainMuller:
```suggestion
    public readonly manifests: readonly string[],
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+    public readonly metadata?: {[key: string]: ResourceConstructMetadata},
+
+    /**
+     * Whether or not the synth command was executed with --stdout.
+     *
+     * @default false
+     */
+    public readonly stdout?: boolean) {
+
+    this._report = new ValidationReport(this.pkg, this.version, this.metadata ?? {}, stdout ?? false);
```
RomainMuller:
```suggestion
    /**
     * Construct metadata of resources in the application.
     */
    public readonly metadata?: {readonly [key: string]: ResourceConstructMetadata} = {},

    /**
     * Whether or not the synth command was executed with --stdout.
     */
    public readonly stdout = false) {

    this._report = new ValidationReport(this.pkg, this.version, this.metadata, stdout);
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  /**
+   * Report emitted by the validation.
+   *
+   * Plugins should interact with this object to generate the report.
+   */
+  public get report(): ValidationReport {
+    return this._report;
+  }
```
RomainMuller: Why not simply rename `this._report` to `public readonly report: ValidationReport`?

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  /**
+   * Logger for the validation.
+   *
+   * Plugins should interact with this object to log messages during validation.
+   */
+  public get logger(): ValidationLogger {
+    return this._logger;
+  }
```
RomainMuller: Same as above... I don't understand why the dynamic accessor is necessary?

If the concern is you want to be 100% certain people don't "cheat" in "pure JS" here... the privates should be `#privateFields` instead of being soft-privates... Then it would make sense to me.

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+ */
+export class ValidationLogger {
+
+  /**
+   * Log a message.
+   */
+  public log(message: string) {
+    console.log(message);
+  }
+}
```
RomainMuller, line 77: Are you somehow preventing use of `console.log`? There might be ways to achieve this...

If not... why not let plug-ins use `console.log` or `console.error` as they see fit? Any particular reason?

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  /**
+   * The resource name.
+   */
+  readonly resourceName: string;
+
+  /**
+   * The locations in its config that pose the violations.
+   */
+  readonly locations: string[];
```
RomainMuller:
```suggestion
  readonly locations: readonly string[];
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  /**
+   * How to fix the recommendation.
+   */
+  readonly fix: string;
+
+  /**
+   * The resources violating this rule.
+   */
+  readonly violatingResources: ValidationViolatingResource[];
```
RomainMuller:
```suggestion
  readonly violatingResources: readonly ValidationViolatingResource[];
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+/**
+ * Validation produced by the validation plugin, in construct terms.
+ */
+export interface ValidationViolationConstructAware extends Omit<ValidationViolation, 'violatingResources'> {
+
+  /**
+   * The constructs violating this rule.
+   */
+  readonly violatingConstructs: ValidationViolatingConstruct[];
```
RomainMuller:
```suggestion
  readonly violatingConstructs: readonly ValidationViolatingConstruct[];
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+ */
+export interface ValidationReportSummary {
+
+  readonly status: ValidationReportStatus;
+
+  readonly plugin: string;
+
+  readonly version: string;
+
+  readonly metadata?: { [key: string]: string };
```
RomainMuller:
```suggestion
  readonly metadata?: { readonly [key: string]: string };
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  /**
+   * Report title.
+   */
+  readonly title: string;
+
+  /**
+   * List of violations in the rerpot.
+   */
+  readonly violations: ValidationViolationConstructAware[];
```
RomainMuller:
```suggestion
  readonly violations: readonly ValidationViolationConstructAware[];
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+export class ValidationReport {
+
+  private readonly violations: ValidationViolationConstructAware[] = [];
+
+  private _summary?: ValidationReportSummary;
+
+  constructor(
+    private readonly pkg: string,
+    private readonly version: string,
+    private readonly metadata: {[key: string]: ResourceConstructMetadata},
```
RomainMuller:
```suggestion
    private readonly metadata: {readonly [key: string]: ResourceConstructMetadata},
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+      recommendation: violation.recommendation,
+      violatingConstructs: violatingConstructs,
+      fix: violation.fix,
+    });
+  }
+
+  /**
+   * Submit the report with a status and additional metadata.
+   */
+  public submit(status: ValidationReportStatus, metadata?: { [key: string]: string }) {
```
RomainMuller:
```suggestion
  public submit(status: ValidationReportStatus, metadata?: { readonly [key: string]: string }) {
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+      ...Object.entries(json.summary.metadata ?? {}),
+    ]));
+
+    if (json.violations) {
+      output.push('');
+      output.push('(Violations)');
+    }
+    for (const violation of json.violations) {
+      const occurrences = violation.violatingConstructs.flatMap(c => c.locations).length;
+      const title = `\x1b[1m\x1b[31m${violation.ruleName} (${occurrences} occurrences)\x1b[0m`;
```
RomainMuller: Nitpick: the ANSI escapes here are not readable... Would be dope to encode them better (maybe with colors-like helpers).

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+ * Utiliy class for loading validation plugins.
+ */
+export class ValidationPlugin {
+
+  /**
+   * Load the validation plugin and create the necessary context for its execution.
+   */
+  public static load(
+    validation: ValidationConfig,
+    manifests: string[],
```
RomainMuller:
```suggestion
    manifests: readonly string[],
```

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+      pkg: validation.package,
+      version: validation.version,
+      class: validation.class,
+      properties: validation.properties,
+      installEnv: validation.installEnv,
+    });
+
+    if (typeof plugin.instance.validate !== 'function') {
+      throw new Error(`Instance of class ${validation.class} is not a validation plugin. Are you sure you specified the correct class?`);
+    }
```
RomainMuller: Might be useful to specify the package name & version, too... there may be multiple plug-ins with the same class name configured...

src/plugins/validation.ts
```diff
@@ -0,0 +1,388 @@
...
+
+    // TODO: parse from manifests
+    const metadata = {};
```
RomainMuller: I guess this is meant to be done?

src/util.ts
```diff
@@ -53,15 +56,59 @@ export async function synthApp(command: string, outdir: string) {
...
     found = true;
   }
 
   if (!found) {
     console.error('No manifests synthesized');
   }
+  return yamlFiles;
+}
+
+export async function validateManifests(manifests: string[], stdout: boolean, validations: ValidationConfig[], pluginManager: PluginManager) {
```
RomainMuller:
```suggestion
export async function validateManifests(manifests: readonly string[], stdout: boolean, validations: readonly ValidationConfig[], pluginManager: PluginManager) {
```

RomainMuller (changes requested):
Looking pretty good, some minor comments...

Generally speaking, I recommend tightening down the contracts you allow plug-in authors to assume by making everything `readonly` unless if needs to be mutated (including arrays and objects). This will signal that mutations on these are not supported and they cannot rely on such side-effects to be upheld (even if they accidentally work at some point).

You might want to also use ES private fields (`#field`) for internal details that you do not want plugins to _be able_ to touch.

src/plugins/validation.ts
iliapolo: Right, waiting on https://github.com/cdk8s-team/cdk8s-core/pull/721. I'll get on it.

src/plugins/validation.ts
iliapolo: Done

src/plugins/validation.ts
iliapolo: 👍

src/plugins/validation.ts
iliapolo: Same

src/plugins/validation.ts
iliapolo: Again just java mannerisms

src/plugins/validation.ts
iliapolo: 👍

src/plugins/_manager.ts
iliapolo: The hot path of this code is for checking our own plugins directory, which we control so we know the structure of. 
Its true that this is also called for local plugins, and may fail in some environments. Im kind of ok with handling those as they come, especially since it was designed to make plugin **development** easier, not **consumption**. 

Now that I think of it though, the local case creates too much diversion in the code. I'm leaning towards getting rid of it altogether, and have plugin developers symlink their plugin into the expected plugins directory, in the expected structure. What do you think?

src/plugins/_manager.ts
iliapolo: Right, switched the order 👍

src/plugins/_manager.ts
iliapolo: `execSync` doesn't have the `shell` option, it actually always runs it in a shell (unlike `spawnSync`).

src/plugins/_manager.ts
iliapolo: I like keeping exception cases rooted by different causes separate, even if they are eventually handled the same.
It makes the code easier to reason about I think, even if only by allowing different comments.

src/plugins/_manager.ts
iliapolo: 👍

src/plugins/_manager.ts
iliapolo: Well damn. Good catch. Added `semver` as a dependency and validated using `semver.Range(v).range`.

src/plugins/_manager.ts
iliapolo: I still have some java mannerisms :)

src/plugins/_manager.ts
iliapolo: Just an omission.

src/plugins/_manager.ts
iliapolo: 👍

src/config.ts
iliapolo: (going to ignore :))

src/cli/cmds/synth.ts
iliapolo: Nice. Thanks.

.projenrc.ts
iliapolo: But I still want to actively prevent other imports, which only this can achieve. Right?

iliapolo: Consider exporting `package.json` as well for consumers to determine which version of the CLI is being used.

iliapolo: We add that if needed later on, not gonna bother with it just now.

src/plugins/_manager.ts
iliapolo: As it turns out, this whole thing is unnecessary because we already encode the plugin version in the path.

src/plugins/validation.ts
iliapolo: Done

src/plugins/_manager.ts
iliapolo: We talked about using the `Validation` interface here to type the `instance` property. 
However, at this point in the code, we don't actually know it is indeed an instance of `Validation`, this is being validated later on. The plugin manager just returns an instance of something, its not aware of what specific plugin invoked it, and its up to the specific plugin to perform its validations and cast it to the concrete type:

https://github.com/cdk8s-team/cdk8s-cli/blob/547114a6e5809bfb8ab100cdd6f2b99126b56698/src/plugins/validation.ts#L365-L371

I'm going to switch this to `unknown`, just so that consumers are forced to check it, even if that checks amounts to casting to `any` themselves.

src/plugins/validation.ts
iliapolo: So like we talked about, I initially did this so that we can control `synth` output for the `--json`. I decided eventually to allow specifying a file where the report will be outputed to in JSON format. 

Doing this via `stdout` was just too big of a change at this time, and we can always add it later if needed. The file is simple and straight forward.

RomainMuller (approved):

PR #668, opened 2022-11-29
fix: `synth` output in `stdout` mode cannot be piped to `kubectl`

### LINKED ISSUE #667: Unable to utilize --stdout flag correctly
When using the `cdk8s synth --stdout` command line the words "Synthesizing application" appear in STDOUT prior to the valid kubernetes manifests. This is caused by [util.ts line 43](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/util.ts#L43). The main issue here is when you expect the following to work:

```
cdk8s synth --stdout | kubectl apply -f -
```

What you wind up with is an errror:

```
error: error parsing STDIN: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context
```

As a work around it is easy enough to do:

```
cdk8s synth --stdout | grep -v 'Synthesizing application' | kubectl apply -f -
```

However it seems easy enough to check `if stdout` and skip the `console.log` entirely. Curious what your thoughts are here, if this change is acceptable I will open a PR shortly.

crazed:
Fixes #667


iliapolo:
@crazed looks good. just need to resolve DCO. See https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/CONTRIBUTING.md#developer-certificate-of-origin-dco

crazed:
> @crazed looks good. just need to resolve DCO. See https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/CONTRIBUTING.md#developer-certificate-of-origin-dco

@iliapolo thanks, first time using commit signing, let me know if i got it right

iliapolo (approved):

PR #890, opened 2023-03-31
feat(import): sync cli `cdk8s import <url>` with `cdk8s.yaml` config file

### LINKED ISSUE #886: Add urls from `cdk8s import <url>` to `cdk8s.yaml`
### Description of the feature or enhancement:

Currently there are two ways of importing crds in cdk8s. A user can mention the url within `cdk8s.yaml` file or run `cdk8s import <url>` command. 
Executing the CLI command does not add the url to the `cdk8s.yaml`. It would be great to have the CLI import being reflected in `cdk8s.yaml` file too.

### Use Case:
This feature would help keep crds that are imported to be consistent between cdk8s.yaml and the CLI import. 

### Proposed Solution:
As part of the cdk8s CLI import code execution, the `cdk8s.yaml` file could be updated with the url for the crd passed in the import command. This url is available as `yargs` values and can be added to the yaml file.
I believe this should be the default behavior when importing crds via CLI. But if otherwise, there can be a flag added to the import command for user to opt in this feature.

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

sumupitchayan:
Fixes #886 

Users can currently import crds in cdk8s in one of two ways: either by adding the url to the imports section in the `cdk8s.yaml` config file or by running the `cdk8s import <url>` command. The issue is that running the CLI import command does not add the url to the config `cdk8s.yaml` file, which could cause confusion to users over which imports they actually have.

This PR addresses this issue by making sure all new imports added with the `cdk8s import <url>` command via the CLI are added to the `cdk8s.yaml` config file.

```yaml
language: typescript
app: node main.js
imports:
  - k8s
```

I added tests in the new `test/import/import-crd.test.ts` file.


src/import/base.ts
```diff
@@ -30,6 +30,11 @@ export interface ImportOptions {
    * k8s imports will add a "Kube" prefix by default.
    */
   readonly classNamePrefix?: string;
+
+  /**
+   * Path that contains the cdk8s.yaml output file.
+   */
+  readonly configFilePath?: string;
```
iliapolo: Not sure I understand where this is being used...our config file path isn't configurable - can you clarify?

src/config.ts
```diff
@@ -32,15 +32,36 @@ export interface Config {
...
+
+  if (!curConfig.imports?.includes(source)) {
+    const importsList = curConfig.imports ?? [];
+    importsList.push(source);
+    let config = {
+      language: curConfig.language,
+      app: curConfig.app,
+      imports: importsList,
+      output: curConfig.output ?? DEFAULTS.output,
+      pluginsDirectory: curConfig.pluginsDirectory ?? DEFAULTS.pluginsDirectory,
```
iliapolo: This is problematic because it would create a file with, for example:

```yaml
pluginsDirectory: /Users/epolon/.cdk8s/plugins
```

Notice the `/Users/epolon/` now makes this file non portable - for example to run in CI builds...

src/config.ts
```diff
@@ -32,15 +32,36 @@ export interface Config {
...
     config = {
       ...config,
-      ...yaml.parse(fs.readFileSync(CONFIG_FILE, 'utf-8')),
+      ...parsedYaml,
     };
   }
   return config;
+}
+
+export async function addImportToConfig(source: string, filePath?: string) {
```
iliapolo: The function is called `addImport`, but it actually does more than that by persisting default values to the file.

Also, I actually don't think we should be writing anything else to the file. It creates a weird experience where the `cdk8s import` command suddenly adds unrelated things to the file. Think about it from a user perspective - would they understand why the file changed in such a way?

Lets simply read the current config, append the import to it, and write back. Then, the function is actually appropriately named :)

test/import/import-cli-config-sync.test.ts
iliapolo, line 1: No need for a separate file for these tests. Just add the tests to the [existing file](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/test/import/import-crd.test.ts).

test/import/import-cli-config-sync.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    await fs.outputFile(path.join(configFilePath, 'cdk8s.yaml'), yaml.stringify(defaultConfig));
+  }
+}
+
+test('reading default config file works', async () => {
+
+  // clean up cdk8s.yaml file, reseting back to template:
+  const defaultConfig = yaml.parse(fs.readFileSync(path.join(configFilePath, 'cdk8s-template.yaml'), 'utf-8'));
+  await fs.outputFile(path.join(configFilePath, 'cdk8s.yaml'), yaml.stringify(defaultConfig));
+  const config1 = readConfigSync(configFilePath);
```
iliapolo: Ah, is this why you added `configFilePath` to the `ImportOptions`? 

In general, whenever you find yourself adding production code solely for the sake of unit tests, think twice, and then another time :). In some *rare* cases its ok, but I don't think we are there.

 Normally, this just means you are testing the wrong thing. For example here you are testing a code path doesn't exist in production, so its not needed.

test/import/import-cli-config-sync.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  await addImportToConfig(jenkinsCRD, configFilePath);
+  const config2 = readConfigSync(configFilePath);
+  expect(config2.imports).toEqual(['k8s', jenkinsCRD]);
+  expect(config2.language).toEqual('typescript');
+  expect(config2.app).toEqual('node main.js');
+
+  // test importing duplicate does not add to config
+  await addImportToConfig(jenkinsCRD, configFilePath);
+  const config3 = readConfigSync(configFilePath);
+  expect(config3).toEqual(config2);
```
iliapolo: This case deserves its own test, where you import the same CRD twice.

test/import/import-cli-config-sync.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  // clean up cdk8s.yaml file, reseting back to template:
+  const defaultConfig = yaml.parse(fs.readFileSync(path.join(configFilePath, 'cdk8s-template.yaml'), 'utf-8'));
+  await fs.outputFile(path.join(configFilePath, 'cdk8s.yaml'), yaml.stringify(defaultConfig));
```
iliapolo: Special cleanup code inside individual tests is a code smell. It means you have a shared resource that shouldn't be shared. Also keep in mind that tests can run in parallel, in which case it can cause race conditions. 

Instead, any mutable resource required by a test should be created by the test itself.

test/import/import-cli-config-sync.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  const config1 = readConfigSync(configFilePath);
+  expect(config1.language).toEqual('typescript');
+  expect(config1.app).toEqual('node main.js');
+  expect(config1.imports?.length == 1).toBeTruthy();
+  expect(config1.imports?.includes('k8s')).toBeTruthy();
+});
+
+test('writing default config file works', async () => {
+  await addImportToConfig(jenkinsCRD, configFilePath);
```
iliapolo: This isn't the right thing to invoke in the test. What if we by mistake remove the call to `addImportToConfig` from the `importDispatch` function? This test would pass, but the functionality will be broken. 

Ideally, the test should invoke `cdk8s import` - just like a user would do. Since its a little awkward, we can invoke `importDispatch` and test its result. This is sufficient because if we remove `importDispatch` from the `cdk8s import` command by mistake, our init tests will fail.

test/import/import-cli-config-sync.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      },
+    },
+  };
+
+  await withTempFixture(crd, async (fixture: string, configFile: string, cwd: string) => {
+    const config1 = readConfigSync(configFile);
+    expect(fixture).not.toEqual(configFile);
+    expect(config1.language).toEqual('typescript');
+    expect(config1.app).toEqual('node main.js');
+    expect(config1.imports).toEqual(['k8s']);
```
iliapolo: What are you asserting on here? that the `readConfigSync` function reads the file correctly? We already have tests for that...no need to re-assert.

test/import/import-cli-config-sync.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  await addImportToConfig(jenkinsCRD, configFilePath);
+  const config3 = readConfigSync(configFilePath);
+  expect(config3).toEqual(config2);
+
+  // clean up cdk8s.yaml file, reseting back to template:
+  const defaultConfig = yaml.parse(fs.readFileSync(path.join(configFilePath, 'cdk8s-template.yaml'), 'utf-8'));
+  await fs.outputFile(path.join(configFilePath, 'cdk8s.yaml'), yaml.stringify(defaultConfig));
+});
+
+test('can import crds from fixtures', async () => {
```
iliapolo: Why do we need this test? What production scenario is this testing?

iliapolo (changes requested):

src/config.ts
```diff
@@ -32,15 +32,42 @@ export interface Config {
...
+    const importsList = curConfig.imports ?? [];
+    importsList.push(source);
+    let config = {
+      ...curConfig,
+      imports: importsList,
+    };
+
+    // just adds the new import if the cdk8s.yaml file already exists
+    // if the cdk8s.yaml does not exist, then it creates it from the defaults
+    if (fs.existsSync(CONFIG_FILE)) {
```
iliapolo: Not sure I understand why is this needed - its already covered in `readConfigSync` - no?

src/config.ts
```diff
@@ -32,15 +32,42 @@ export interface Config {
...
+      ...parsedYaml,
     };
   }
   return config;
+}
+
+export async function addImportToConfig(source: string) {
+  let curConfig = readConfigSync();
+
+  if (!curConfig.imports?.includes(source)) {
```
iliapolo:
```suggestion
  const currImports = curConfig.imports ?? [];
  if (!currImports.includes(source)) {
```

And then push directory to `currImports` below

src/config.ts
```diff
@@ -32,15 +32,42 @@ export interface Config {
 const DEFAULTS: Config = {
   output: 'dist',
   pluginsDirectory: path.join(os.homedir(), '.cdk8s', 'plugins'),
+  imports: ['k8s'],
```
iliapolo: How come?

test/import/import-crd.test.ts
```diff
@@ -465,3 +467,52 @@ test('given a prefix, we can import two crds with the same group id', async () =
...
+    // creates default cdk8s.yaml file at root of the project directory
+    const defaultConfigPath = path.join(__dirname, 'cdk8s-template.yaml');
+    const defaultConfig = yaml.parse(fs.readFileSync(defaultConfigPath, 'utf-8'));
+    await fs.outputFile('cdk8s.yaml', yaml.stringify(defaultConfig));
+  });
+
+  afterAll(async () => {
+    // deletes the cdk8s.yaml file created at the root of the project directory and crd-imports folder
+    await fs.unlink('cdk8s.yaml');
+    await fs.rmSync(path.join(__dirname, 'crd-imports'), { recursive: true, force: true });
```
iliapolo: `await` not need on a sync command.

```suggestion
    fs.rmSync(path.join(__dirname, 'crd-imports'), { recursive: true, force: true });
```

test/import/cdk8s-template.yaml
```diff
@@ -0,0 +1,4 @@
+language: typescript
```
iliapolo: Prefer to use create this file in the test itself - if its simple enough. That is, write this file to disk from within the test.

test/import/import-crd.test.ts
```diff
@@ -465,3 +467,52 @@ test('given a prefix, we can import two crds with the same group id', async () =
...
 });
+
+describe('cdk8s.yaml file', () => {
+  const jenkinsCRD: ImportSpec = {
+    source: 'https://raw.githubusercontent.com/jenkinsci/kubernetes-operator/master/deploy/crds/jenkins.io_jenkins_crd.yaml',
+  };
+
+  const importOptions: ImportOptions = {
+    targetLanguage: Language.TYPESCRIPT,
+    outdir: path.join(__dirname, 'crd-imports'),
```
iliapolo: Never use `__dirname` is a working directory in tests - it prevents them from running in parallel (imagine two tests, one deleting this dir and one reading from it...). Instead, each test should create a temporary working directory via `fs.mkdtempSync`

test/import/import-crd.test.ts
```diff
@@ -465,3 +467,52 @@ test('given a prefix, we can import two crds with the same group id', async () =
...
+  };
+
+  beforeEach(async () => {
+    // creates default cdk8s.yaml file at root of the project directory
+    const defaultConfigPath = path.join(__dirname, 'cdk8s-template.yaml');
+    const defaultConfig = yaml.parse(fs.readFileSync(defaultConfigPath, 'utf-8'));
+    await fs.outputFile('cdk8s.yaml', yaml.stringify(defaultConfig));
```
iliapolo: Again the `cdk8s.yaml` file is now a global shared resource...bad for parallelism. Instead, create a temporary directory and use `process.chdir` (parallel tests will run in separate processes so it is safe)

test/import/import-crd.test.ts
```diff
@@ -465,3 +467,52 @@ test('given a prefix, we can import two crds with the same group id', async () =
...
+  });
+
+  afterAll(async () => {
+    // deletes the cdk8s.yaml file created at the root of the project directory and crd-imports folder
+    await fs.unlink('cdk8s.yaml');
+    await fs.rmSync(path.join(__dirname, 'crd-imports'), { recursive: true, force: true });
+  });
+
+  test('can be read by default', async () => {
```
iliapolo: So now that `readConfigSync` is left untouched - why do you need this?

iliapolo (changes requested):

test/import/import-crd.test.ts
iliapolo: Super dangerous code because it operates on the repository directory - one small bug or mistake during development and it can wipe out your own code...(speaking from experience). Another reason why tests should only operate on temporary directories.

src/config.ts
sumupitchayan: Just saw that the default template has k8s as its only import, I guess it's unnecessary though so I can delete it.

Would it ever be the case that someone deletes the cdk8s.yaml file?

test/import/import-crd.test.ts
sumupitchayan: I guess I don't need it - I had it originally just to make sure that reading the default cdk8s.yaml file worked. If it makes more sense to delete it I can remove this test

iliapolo (approved):

shinebayar-g:
Hi @sumupitchayan , @iliapolo looks like this was the root cause of https://github.com/cdk8s-team/cdk8s/issues/1442

cdk8s import command is broken since release 2.8.0

PR #896, opened 2023-04-06
fix(init-typescript): `cdk8s synth` fails if `ts-node` is not globally installed

### LINKED ISSUE #1273: ts-node: command not found
### Description of the bug:

`cdk8s synth` fails with the error message below. It looks like `ts-node` needs to be globally installed, b/c a `npm install -g ts-node` resolves the issue. Based on the getting started guide, I did assume that no additional install steps are necessary. When I noticed that `ts-node` was included in the project scaffolding as a devDependency it did surprise me that the CLI is not using the ts-node version bundled with the project.

- `cdk8s` version: 2.2.10

### Reproduction Steps:

1. `npm install -g cdk8s-cli`
2. `cdk8s init typescript-app`
3. `npm run compile && cdk8s synth`

### Error Log:

```
Synthesizing application
/bin/sh: ts-node: command not found
Error: command "ts-node main.ts " at <redact> returned a non-zero exit code 127
    at ChildProcess.<anonymous> (/opt/homebrew/lib/node_modules/cdk8s-cli/lib/util.js:51:27)
    at Object.onceWrapper (node:events:627:26)
    at ChildProcess.emit (node:events:512:28)
    at ChildProcess._handle.onexit (node:internal/child_process:293:12)
```

### Environment:

  - **Framework Version**: "typescript": "4.9.5"
  - **OS**: macOS 13.3 (M1)

### Other:

---

This is :bug: Bug Report

shinebayar-g:
Fixes cdk8s-team/cdk8s#1273

### DIFF
--- a/templates/typescript-app/cdk8s.yaml
+++ b/templates/typescript-app/cdk8s.yaml
@@ -1,4 +1,4 @@
 language: typescript
-app: ts-node main.ts
+app: npx ts-node main.ts
 imports:
   - k8s


templates/typescript-app/cdk8s.yaml
```diff
@@ -1,4 +1,4 @@
 language: typescript
-app: ts-node main.ts
```
iliapolo, line 2: How about we just do `npx ts-node main.ts`? Normally our direction of indirection (funny) is `package.json -> cdk8s.yaml`, not the other way around :)

shinebayar-g: npx will still globally install the package and I think that's unnecessary dependency since it's already in local `node_modules`?

Invoking `npm run synth` will utilize local dependency and `cdk8s synth` command stays the same across languages. If it's not a desired approach users could just use `npm run synth` (from current package.json: `"synth": "cdk8s synth",`) instead of `cdk8s synth`. I'm guessing that was probably the motivation behind why this npm script is defined?

iliapolo: @shinebayar-g `npx` is included with the node distribution, so its not an additional dependency. This is also the way we do it in the aws-cdk: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/init-templates/app/typescript/cdk.template.json#L2

`npx` will also utilize the local `node_modules` dependencies and has the benefit of installing it on-demand if its missing. I'd like to keep the `cdk8s.yaml` file self contained and clean of `package.json` scripts.

shinebayar-g: > npx will also utilize the local node_modules dependencies and has the benefit of installing it on-demand if its missing.

Oh I completely missed this detail. Just checked the behavior, you're right. I always thought it installs its own dependency with own execution environment, that was false. It prefers local first.

I'll update the PR.

iliapolo (approved):
Sweet. Thanks for taking this 👍

iliapolo:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.x|[](https://github.com/cdk8s-team/cdk8s-cli/pull/897)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

PR #1020, opened 2023-06-19
fix(go): `cdk8s init` fails go golang app when cli is installed via `yarn`

iliapolo:
We are seeing the following [failures](https://github.com/cdk8s-team/cdk8s/actions/runs/5306513954/jobs/9604183388?pr=1335):

```console
Importing k8s v1.25.0...
Importing resources, this may take a few moments...
go: errors parsing go.mod:
/tmp/tmp.ullrWKlhR2/test/go.mod:7:2: require github.com/aws/jsii-runtime-go: version "v5.1.2" invalid: should be v0 or v1, not v5
````

When the CLI is installed via Yarn, the version of `jsii` that ends up in the root of the `node_modules` tree is the one coming transitively from the [jsii-srcmak](https://github.com/cdklabs/jsii-srcmak/blob/main/package.json#L61), which is `5.1.1` (see https://github.com/cdklabs/jsii-srcmak/pull/1118). The CLI incorrectly uses that as the version of `jsii-runtime-go` and fails. 

This PR changes the jsii version detection logic to look at the version of `jsii-pacmak` instead of `jsii`, now that they have different version lines. Also added integration tests for `cdk8s init` when the CLI is installed via Yarn.

Note that for the NPM case, which is the only one we tested until now, the `jsii` that ends up in the root is the one coming transitively from `jsii-pacmak`, which still remains `1.84.0`.


RomainMuller (approved):

iliapolo:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|1.x|[](https://github.com/cdk8s-team/cdk8s-cli/pull/1021)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

PR #1042, opened 2023-06-25
fix(import): failure when multiple versions of the same CRD exist in different obejcts

iliapolo:
When a CRD has multiple objects that represent the same *kind* but for different versions, `cdk8s import` will fail with a vague "already exists" error.

For example:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: tenants.capsule.clastix.io
spec:
  group: capsule.clastix.io
  names:
    kind: Tenant
    listKind: TenantList
    plural: tenants
    shortNames:
      - tnt
    singular: tenant
  scope: Cluster
  versions:
    - name: v1beta1
      schema:
        openAPIV3Schema: ...
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: tenants.capsule.clastix.io
spec:
  group: capsule.clastix.io
  names:
    kind: Tenant
    listKind: TenantList
    plural: tenants
    shortNames:
      - tnt
    singular: tenant
  scope: Cluster
  versions:
    - name: v1beta2
      schema:
        openAPIV3Schema: ...

```

Note that if the two versions are defined within the same object, import will succeed. 

For example:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: tenants.capsule.clastix.io
spec:
  group: capsule.clastix.io
  names:
    kind: Tenant
    listKind: TenantList
    plural: tenants
    shortNames:
      - tnt
    singular: tenant
  scope: Cluster
  versions:
    - name: v1beta1
      schema:
        openAPIV3Schema: ...
    - name: v1beta2
      schema:
        openAPIV3Schema: ...
```

This is actually a followup to https://github.com/cdk8s-team/cdk8s-cli/issues/387.


src/import/base.ts
```diff
@@ -72,7 +72,7 @@ export abstract class ImportBase {
 
     for (const module of modules) {
       // output the name of the imported resource
-      console.log(module.origName);
+      // console.log(module.origName);
```
vinayak-kukreja: Any reason to have `console.log` around?

iliapolo: Good catch. It wasn't supposed to be commented out. Fixed.

vinayak-kukreja (approved):

PR #1195, opened 2023-08-07
feat: helm synthesis

vinayak-kukreja:
Implementation for [Helm Synthesis RFC](https://github.com/cdk8s-team/cdk8s/issues/1251).

Resolves https://github.com/cdk8s-team/cdk8s-team-internal/issues/53


projenrc/integ.ts
```diff
@@ -23,7 +27,8 @@ export function addIntegTests(project: typescript.TypeScriptProject) {
 
   const templatesDir = path.join(__dirname, '..', 'templates');
   for (const template of fs.readdirSync(templatesDir)) {
-    if (fs.statSync(path.join(templatesDir, template)).isDirectory()) {
+    if (fs.statSync(path.join(templatesDir, template)).isDirectory() &&
```
iliapolo: These templates shouldn't be inside the `templates` directory because they are not init templates. Either put them in a different directory, or better yet, remove them and dynamically create the files in the code? all the files there seem very basic.

src/cli/cmds/synth.ts
```diff
@@ -18,7 +23,16 @@ class Command implements yargs.CommandModule {
     .option('stdout', { type: 'boolean', required: false, desc: 'Write synthesized manifests to STDOUT instead of the output directory', alias: 'p' })
     .option('plugins-dir', { default: config.pluginsDirectory, required: false, desc: 'Directory to store cdk8s plugins.' })
     .option('validate', { type: 'boolean', default: true, required: false, desc: 'Apply validation plugins on the resulting manifests (use --no-validate to disable)' })
-    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' });
+    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' })
+    .option('format', {
+      default: config.format,
+      required: false,
+      desc: 'Synthesis format for Kubernetes manifests.',
+      choices: ['cdk8s', 'helm'],
```
iliapolo:
```suggestion
      choices: ['plain', 'helm'],
```

`cdk8s` is not a format, we produce "plain" manifests by default.

src/cli/cmds/synth.ts
```diff
@@ -18,7 +23,16 @@ class Command implements yargs.CommandModule {
...
+      type: 'string',
+    })
+    .option('chart-api-version', { default: config.helmSynthConfig?.chartApiVersion, required: false, desc: 'Chart API version of helm chart.' })
+    .option('chart-version', { required: false, desc: 'Chart version of helm chart.' });
```
iliapolo: Mention this is required if format is `helm`.

src/cli/cmds/synth.ts
```diff
@@ -18,7 +23,16 @@ class Command implements yargs.CommandModule {
...
-    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' });
+    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' })
+    .option('format', {
+      default: config.format,
+      required: false,
+      desc: 'Synthesis format for Kubernetes manifests.',
+      choices: ['cdk8s', 'helm'],
+      type: 'string',
+    })
+    .option('chart-api-version', { default: config.helmSynthConfig?.chartApiVersion, required: false, desc: 'Chart API version of helm chart.' })
```
iliapolo: Mention what the default value will be.

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
 }
 
 
+async function helmSynthesis(apiVersion: string, chartVersion: string, outdir: string) {
+  const TEMPLATE_WITHOUT_CRDS = 'helm-chart-without-crds';
+  const TEMPLATE_WITH_CRDS = 'helm-chart-with-crds';
+
+  const pkgroot = path.join(__dirname, '..', '..', '..');
+  const templatesDir = path.join(pkgroot, 'templates');
```
iliapolo: All of these can actually live outside this function as global constants.

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
 }
 
 
+async function helmSynthesis(apiVersion: string, chartVersion: string, outdir: string) {
```
iliapolo:
```suggestion
async function createHelmScaffolding(apiVersion: string, chartVersion: string, outdir: string) {
```

src/cli/cmds/synth.ts
```diff
@@ -29,6 +43,9 @@ class Command implements yargs.CommandModule {
     const validate = argv.validate;
     const pluginsDir = argv.pluginsDir;
     const reportFile = argv.validationReportsOutputFile;
+    const format = argv.format;
```
iliapolo: Shouldn't we validate this can only be `plain` or `helm` somewhere? what is preventing me from passing "bla"?

src/cli/cmds/synth.ts
```diff
@@ -29,6 +43,9 @@ class Command implements yargs.CommandModule {
...
+    const chartApiVersion = argv.chartApiVersion;
```
iliapolo: Same here - where is the validation that this can only be `v1` or `v2`?

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+  const substituteValues = {
+    apiVersion: apiVersion,
+    version: chartVersion,
+    app: path.basename(path.resolve()),
+  };
+
+  // Helm chart structure without CRDs support
+  if (apiVersion === HelmChartApiVersion.V1) {
+
+    if (config.imports && config.imports.length > 0) {
```
iliapolo: This is not the correct check because our normal `k8s` imports are also part of this, but they are not CRDs. Make sure we have a test for this scenario.

src/config.ts
```diff
@@ -20,28 +20,58 @@ export interface ValidationConfig {
   readonly properties?: { [key: string]: any };
 }
 
+export enum SynthesisFormat {
+  'CDK8s' = 'cdk8s',
+  'HELM' = 'helm',
```
iliapolo:
```suggestion
  PLAIN = 'plain',
  HELM = 'helm',
```

src/config.ts
```diff
@@ -20,28 +20,58 @@ export interface ValidationConfig {
...
+}
+
+export enum HelmChartApiVersion {
+  'V1' = 'v1',
+  'V2' = 'v2',
```
iliapolo:
```suggestion
  V1 = 'v1',
  V2 = 'v2',
```

src/config.ts
```diff
@@ -20,28 +20,58 @@ export interface ValidationConfig {
...
+
 export interface Config {
   readonly app?: string;
   readonly language?: Language;
   readonly output?: string;
   readonly imports?: string[];
   readonly pluginsDirectory?: string;
   readonly validations?: string | ValidationConfig[];
+  readonly format?: SynthesisFormat;
+  readonly helmSynthConfig?: HelmSynthesis;
```
iliapolo: This makes the config file kind of cumbersome. How about we have:

```yaml
language: typescript
...
synth:
  format: helm
  chartVersion: 0.0.1
  chartApiVersion: v2
```

?

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+    templatePath = path.join(templatesDir, TEMPLATE_WITH_CRDS);
+  }
+
+  try {
+    await sscaff(templatePath, outdir, substituteValues);
+  } catch (error) {
+    throw new Error(`An error occurred during Helm chart creation: ${error}`);
+  }
+
+  await addCrdsToHelmChart(outdir);
```
iliapolo: Shouldn't this be invoked only when CRDs are supported? i.e v2?

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+          parsedToJson.kind === CRD_KIND &&
+          parsedToJson.metadata &&
+          parsedToJson.metadata.name
+        ) {
+          filename = parsedToJson.metadata.name;
+        }
+      }
+
+      if (!filename) {
+        filename = `CRD-${randomUUID}`;
```
iliapolo: So every time we synthesize the filename gets changed? :)

We can't do this. The filename of the crd should be somehow derived from the URL. If all else fails, a hash of the URL will suffice.

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+  } catch (error) {
+    throw new Error(`An error occurred during Helm chart creation: ${error}`);
+  }
+
+  await addCrdsToHelmChart(outdir);
+}
+
+async function addCrdsToHelmChart(chartDir: string) {
+  try {
+    const CRD_KIND = 'CustomResourceDefinition';
```
iliapolo: It is unconventional to use UPPERCASE inside functions...

```suggestion
    const crdKind = 'CustomResourceDefinition';
```

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+      }
+
+      manifestFiles.push(manifest);
+    }
+  }
+
+  return manifestFiles;
+}
+
+export function parseImports(spec: string): ImportSpec {
```
iliapolo: I suppose this is left over? Because I see you opted to `export` the already existing implementation in `import.ts`.

src/config.ts
```diff
@@ -20,28 +20,58 @@ export interface ValidationConfig {
...
+
+  if (config.format === SynthesisFormat.HELM && config.helmSynthConfig && !config.helmSynthConfig.chartApiVersion) {
+    config = {
+      ...config,
+      helmSynthConfig: {
+        ...config.helmSynthConfig,
+        chartApiVersion: HelmChartApiVersion.V2,
+      },
+    };
+  }
```
iliapolo: This looks like you're setting the default value of `chartApiVersion`, but you're only doing so if `config.helmSynthConfig` is present - why? The most common usage would actually be when the config file doesn't include this, and the `chart-version` is passed via command line. Also note that your default value actually overrides any user defined one...

```suggestion
  if (config.format === SynthesisFormat.HELM) {
    config = {
      ...config,
      helmSynthConfig: {
        chartApiVersion: HelmChartApiVersion.V2,
        ...(config.helmSynthConfig ?? {}),
      },
    };
  }
```

src/util.ts
```diff
@@ -9,7 +9,7 @@ import * as fs from 'fs-extra';
 import * as yaml from 'yaml';
 import { ValidationConfig } from './config';
 import { PluginManager } from './plugins/_manager';
-import { ValidationPlugin, ValidationContext, ValidationReport, Validation } from './plugins/validation';
+import { Validation, ValidationContext, ValidationPlugin, ValidationReport } from './plugins/validation';
```
iliapolo: How come the order here changes? Do you have a different VSCode setting or something?

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+    const e = er as any;
+    throw new Error(`error during project initialization: ${e.stack}\nSTDOUT:\n${e.stdout?.toString()}\nSTDERR:\n${e.stderr?.toString()}`);
+  }
+}
+
+async function downloadCrds(): Promise<string[]> {
+  const manifestFiles = [];
+
+  if (config.imports) {
+    for (const crd of config.imports) {
```
iliapolo: You should ignore `k8s` imports here.

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+      let importSpec;
+      try {
+        importSpec = parseImports(crd);
+      } catch (error) {
+        throw new Error(`I caught you: ${error}`);
+      }
```
iliapolo: I guess it was for debugging? 

```suggestion
      const importSpec = parseImports(crd);
```

src/cli/cmds/synth.ts
```diff
@@ -75,4 +116,129 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+      for (const doc of parsed) {
+        const parsedToJson = doc.toJSON();
+        if (
+          parsedToJson.kind === CRD_KIND &&
+          parsedToJson.metadata &&
+          parsedToJson.metadata.name
+        ) {
+          filename = parsedToJson.metadata.name;
+        }
+      }
```
iliapolo: Why do we need all these? Why not just put the downloaded file in the `crds` directory as is? and derive the filename from the URL? 

Or at least we should reuse the `ImportCustomResourceDefinition.fromSpec` function to perform the download and collect the CRDs.

iliapolo (changes requested):

src/cli/cmds/synth.ts
vinayak-kukreja: That is a good callout. I thought yargs would take care of it: https://yargs.js.org/docs/#api-reference-choiceskey-choices. But I do not see any doc saying that this would throw an error, would test that. 

But, I believe a custom error would be better for users here. Will add this.

src/cli/cmds/synth.ts
vinayak-kukreja: Makes sense. Will test this and make updates.

src/config.ts
vinayak-kukreja: Looks cleaner, thanks Eli 😄

src/cli/cmds/synth.ts
vinayak-kukreja: Yes, I was throwing an error earlier in code if it was v1 and user had crds. I will refactor it to make it more clear.

src/cli/cmds/synth.ts
vinayak-kukreja: I see, taking a look how we can get that :)

src/cli/cmds/synth.ts
vinayak-kukreja: TIL. Thank you :)

src/util.ts
vinayak-kukreja: Yeah, looks like this is a linter thing that sorts imports. Will fix it.

src/cli/cmds/synth.ts
vinayak-kukreja: Lol, yes, sorry about that. 😶‍🌫️

src/cli/cmds/synth.ts
```diff
@@ -38,6 +58,34 @@ class Command implements yargs.CommandModule {
       fs.rmSync(outdir, { recursive: true, force: true });
     }
 
+    if (format && (format != SynthesisFormat.PLAIN && format != SynthesisFormat.HELM)) {
+      throw new Error(`You need to specify synthesis format either as ${SynthesisFormat.PLAIN} or ${SynthesisFormat.HELM} but received: ${format}`);
+    }
```
iliapolo: Actually, this is handled by yargs (because of the `choices` property):

```console
/Users/epolon/dev/src/github.com/cdk8s-team/cdk8s-cli/bin/cdk8s synth --format asdads
Invalid values:
  Argument: format, Given: "asdads", Choices: "plain", "helm"
```

This is good enough, no need for a custom message (not to mention if will never get here :))

src/cli/cmds/synth.ts
```diff
@@ -38,6 +58,34 @@ class Command implements yargs.CommandModule {
...
+
+    if (chartApiVersion && (chartApiVersion != HelmChartApiVersion.V1 && chartApiVersion != HelmChartApiVersion.V2)) {
+      throw new Error(`You need to specify helm chart api version either as ${HelmChartApiVersion.V1} or ${HelmChartApiVersion.V2} but received: ${chartApiVersion}`);
+    }
```
iliapolo, line 63: Not needed if you add `choices` to the flag definition

src/cli/cmds/synth.ts
```diff
@@ -38,6 +58,34 @@ class Command implements yargs.CommandModule {
...
+
+    if (format === SynthesisFormat.HELM && !chartVersion) {
+      throw new Error('You need to specify the \'--chart-version\' when the \'--format\' is set as \'helm\'.');
```
iliapolo:
```suggestion
      throw new Error('You need to specify \'--chart-version\' when \'--format\' is set as \'helm\'.');
```

src/cli/cmds/synth.ts
```diff
@@ -38,6 +58,34 @@ class Command implements yargs.CommandModule {
...
+    }
+
+    if (chartVersion && !semver.valid(chartVersion)) {
+      throw new Error(`The value specified for '--chart-version': ${chartVersion} does not follow SemVer-2(https://semver.org/).`);
```
iliapolo, line 70: Is this a helm requirement?

src/cli/cmds/synth.ts
```diff
@@ -38,6 +58,34 @@ class Command implements yargs.CommandModule {
...
+    }
+
+    if (stdout && format === SynthesisFormat.HELM) {
+      throw new Error('Helm format synthesis does not support \'stdout\'. Please use \'outdir\' instead.');
+    }
+
+    if (format === SynthesisFormat.PLAIN && (chartApiVersion || chartVersion || (chartApiVersion && chartVersion))) {
+      throw new Error('You need to specify \'--format\' as \'helm\' when \'--chart-version\' and/or \'--chart-api-version\' is set.');
+    }
```
iliapolo: Split this out to 3 separate errors, each mentioning the specific flag used.

src/cli/cmds/synth.ts
```diff
@@ -38,6 +58,34 @@ class Command implements yargs.CommandModule {
...
+
+    if (chartApiVersion === HelmChartApiVersion.V1 && crdsArePresent(config.imports)) {
+      throw new Error(`CRDs are not supported with '--chart-api-version': '${HelmChartApiVersion.V1}'. Please use '--chart-api-version': '${HelmChartApiVersion.V2}' for using CRDs.`);
```
iliapolo:
```suggestion
      throw new Error(`Your application uses CRDs, which are not supported with '--chart-api-version': '${HelmChartApiVersion.V1}'. Please either use '--chart-api-version': '${HelmChartApiVersion.V2}' or remove the CRDs from your cdk8s.yaml configuration file`);
```

src/cli/cmds/synth.ts
```diff
@@ -18,7 +26,16 @@ class Command implements yargs.CommandModule {
...
-    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' });
+    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' })
+    .option('format', {
+      default: config.synth!.format,
+      required: false,
+      desc: 'Synthesis format for Kubernetes manifests. The default synthesis format is plain kubernetes manifests.',
+      choices: ['plain', 'helm'],
+      type: 'string',
+    })
+    .option('chart-api-version', { default: config.synth!.chartApiVersion, required: false, desc: 'Chart API version of helm chart. The default value would be \'v2\' api version when synthesis format is helm. There is no default set when synthesis format is plain.' })
```
iliapolo: Add `choices` here

src/cli/cmds/synth.ts
```diff
@@ -18,7 +26,16 @@ class Command implements yargs.CommandModule {
     .option('stdout', { type: 'boolean', required: false, desc: 'Write synthesized manifests to STDOUT instead of the output directory', alias: 'p' })
     .option('plugins-dir', { default: config.pluginsDirectory, required: false, desc: 'Directory to store cdk8s plugins.' })
     .option('validate', { type: 'boolean', default: true, required: false, desc: 'Apply validation plugins on the resulting manifests (use --no-validate to disable)' })
-    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' });
+    .option('validation-reports-output-file', { required: false, desc: 'File to write a JSON representation of the validation reports to' })
+    .option('format', {
+      default: config.synth!.format,
```
iliapolo: Avoid using `!` unless absolutely necessary. Here - I believe it isn't? (make sure to address other places that use this as well)

src/config.ts
```diff
@@ -12,36 +12,67 @@ export interface ImportSpec {
...
       ...config,
       ...yaml.parse(fs.readFileSync(CONFIG_FILE, 'utf-8')),
     };
   }
+
+  if (config.synth?.format === SynthesisFormat.HELM) {
+    config = {
+      ...config,
+      synth: {
+        chartApiVersion: HelmChartApiVersion.V2,
```
iliapolo: Looks like this default is only applied if the user specified `format: helm` in the config file, but what if the config file is left empty, and the user executed this command:

```console
/Users/epolon/dev/src/github.com/cdk8s-team/cdk8s-cli/bin/cdk8s synth --format helm --chart-version 0.0.1
```

In this case, the default value for `chartApiVersion` is undefined, and indeed, the Chart.yaml file looks like this:

```yaml
apiVersion: "undefined"
name: "cdk8s-helm"
version: "0.0.1"
description: Generated chart for cdk8s-helm
type: application
```

It also makes it ignore CRDs in this case. 

Make sure you add tests for all these cases...

src/cli/cmds/synth.ts
```diff
@@ -74,5 +130,79 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+  const substituteValues = {
+    apiVersion: apiVersion,
+    version: chartVersion,
+    app: path.basename(path.resolve()),
+  };
+
+  try {
+    await sscaff(tempHelmStructure, outdir, substituteValues);
+  } catch (error) {
+    throw new Error(`An error occurred during Helm chart creation: ${error}`);
```
iliapolo: No need - what context / information are you adding here that isn't available in the original error and stacktrace?

src/cli/cmds/synth.ts
```diff
@@ -74,5 +130,79 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+  const tempHelmStructure = createFolderStructure();
+
+  const substituteValues = {
+    apiVersion: apiVersion,
+    version: chartVersion,
+    app: path.basename(path.resolve()),
+  };
+
+  try {
+    await sscaff(tempHelmStructure, outdir, substituteValues);
```
iliapolo, line 143: Does `scaff` also cleanup its source directory? i.e `tempHelmStructure`? If not - we should delete it after its done.

src/cli/cmds/synth.ts
```diff
@@ -74,5 +130,79 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+        const manifest = await downloadCrds(source);
+        const filename = deriveFileName(source);
+
+        fs.outputFileSync(path.join(chartDir, 'crds', `${filename}.yaml`), manifest);
+      }
+    }
+  } catch (er) {
+    const e = er as any;
+
+    throw new Error(`Error during adding custom resource definition to helm chart folder: ${e.stack}\nSTDOUT:\n${e.stdout?.toString()}\nSTDERR:\n${e.stderr?.toString()}. `);
```
iliapolo: No need - just let the original error bubble up. There's not much information being added here.

src/cli/cmds/synth.ts
```diff
@@ -74,5 +130,79 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+  const manifest = devUrl ? await download(devUrl): await download(url);
+
+  return manifest;
+}
+
+async function addCrdsToHelmChart(chartDir: string) {
+  try {
+    if (config.imports) {
+      for (const imprt of config.imports) {
+        if (isK8sImport(imprt)) {
```
iliapolo: Apply a `.filter` on `config.import` instead - it reduces the number of branches in the code, which makes it slightly easier to read.

src/cli/cmds/synth.ts
```diff
@@ -74,5 +130,79 @@ async function fetchValidations(): Promise<ValidationConfig[] | undefined> {
...
+    if (config.synth?.chartApiVersion === HelmChartApiVersion.V2 && crdsArePresent(config.imports)) {
+      fs.mkdirSync(path.join(root, 'crds'));
+    }
+
+    return root;
+  }
+}
+
+async function downloadCrds(url: string) {
+  const devUrl = matchCrdsDevUrl(url);
```
iliapolo: This deserves a test. Also - can you to reuse the `matchImporter` function instead? Because if we ever support another CRD source, we will now have two places we would have to handle it in.

src/util.ts
```diff
@@ -255,4 +259,68 @@ export interface SynthesizedApp {
...
+    const lastIndexOfAt = devUrl.lastIndexOf('@');
+    filename = devUrl.slice(lastIndexOfSlash+1, lastIndexOfAt);
+  } else {
+    const lastIndexOfSlash = url.lastIndexOf('/');
+    const lastIndexOfYaml = url.lastIndexOf('.yaml');
+    filename = url.slice(lastIndexOfSlash+1, lastIndexOfYaml);
+  }
+
+  if (!filename) {
+    filename = createHash('sha256');
```
iliapolo: The hash needs to be comprised out of the URL...i.e:

```ts
hash = createHash('sha256');
hash.update(url)
filename = hash.digest('hex')
```

Also note that right now, the function return a `Hash` instance in this path, not a `string`...
You would have caught this if you'd mention the return value of the function in the signature :)

src/util.ts
```diff
@@ -255,4 +259,68 @@ export interface SynthesizedApp {
...
+
+export function crdsArePresent(imprts: string[] | undefined) {
+  if (!imprts) {
+    return false;
+  }
+  if (imprts.length === 0 || (imprts.length === 1 && isK8sImport(imprts[0]))) {
+    return false;
+  }
+
+  return true;
```
iliapolo: This would return true for:

```yaml
- k8s@1.22
- k8s@1.23
```

Which is incorrect.

```suggestion
return (imprts ?? []).some(i => !isK8sImport(i));
```

iliapolo (changes requested):

src/cli/cmds/synth.ts
vinayak-kukreja: Yes. Here: https://helm.sh/docs/topics/charts/#charts-and-versioning

src/cli/cmds/synth.ts
vinayak-kukreja: I was just adding it to make it clear that this particular operation failed. We can know that from the stack trace too, will update it :)

src/util.ts
vinayak-kukreja: Wow, thanks Eli, TIL 🙌

src/cli/cmds/synth.ts
vinayak-kukreja: Adding here since removing `choices` from builder so that we have validations and defaults in handler.

iliapolo:
Converting to draft to wait for AppSec sign-off

src/cli/cmds/synth.ts
```diff
@@ -13,31 +24,72 @@ class Command implements yargs.CommandModule {
...
       throw new Error('\'--output\' and \'--stdout\' are mutually exclusive. Please only use one.');
     }
 
     if (outdir) {
       fs.rmSync(outdir, { recursive: true, force: true });
     }
 
+    if (format != SynthesisFormat.PLAIN && format != SynthesisFormat.HELM) {
+      throw new Error(`You need to specify synthesis format either as ${SynthesisFormat.PLAIN} or ${SynthesisFormat.HELM} but received: ${format}`);
+    }
```
iliapolo, line 59: I believe this is dead code because this validation already happens thanks to the `choices` option...double check? :)

src/cli/cmds/synth.ts
iliapolo: I see you left `choices` in the builder, which is fine - did you mean to do something else here? As of now, this is also dead code.

src/cli/cmds/synth.ts
```diff
@@ -13,31 +24,72 @@ class Command implements yargs.CommandModule {
...
+      throw new Error(`The value specified for '--chart-version': ${chartVersion} does not follow SemVer-2(https://semver.org/).`);
+    }
+
+    if (stdout && format === SynthesisFormat.HELM) {
+      throw new Error('Helm format synthesis does not support \'stdout\'. Please use \'outdir\' instead.');
+    }
+
+    if (format === SynthesisFormat.PLAIN && (chartApiVersion && chartVersion)) {
+      throw new Error('You need to specify \'--format\' as \'helm\' when \'--chart-version\' and \'--chart-api-version\' is set.');
+    }
```
iliapolo: No need. This is covered in the 2 dedicated validations you have below.

src/cli/cmds/synth.ts
```diff
@@ -13,31 +24,72 @@ class Command implements yargs.CommandModule {
...
+    if (format === SynthesisFormat.PLAIN && chartApiVersion) {
+      throw new Error('You need to specify \'--format\' as \'helm\' when \'--chart-api-version\' is set.');
+    }
+
+    if (format === SynthesisFormat.PLAIN && chartVersion) {
+      throw new Error('You need to specify \'--format\' as \'helm\' when \'--chart-version\' is set.');
+    }
+
+    if (chartApiVersion === HelmChartApiVersion.V1 && crdsArePresent(config?.imports)) {
+      throw new Error(`Your application uses CRDs, which are not supported with '--chart-api-version': '${HelmChartApiVersion.V1}'. Please either use '--chart-api-version': '${HelmChartApiVersion.V2}' or remove the CRDs from your cdk8s.yaml configuration file`);
```
iliapolo:
```suggestion
      throw new Error(`Your application uses CRDs, which are not supported when '--chart-api-version' is set to ${HelmChartApiVersion.V1}. Please either set '--chart-api-version' to ${HelmChartApiVersion.V2}, or remove the CRDs from your cdk8s.yaml configuration file`);
```

src/cli/cmds/synth.ts
```diff
@@ -53,26 +105,98 @@ class Command implements yargs.CommandModule {
...
+      type: 'application',
+    };
+
+    fs.outputFileSync(path.join(root, CHART_YAML_FILE), yaml.stringify(chartYamlFile));
+
+    const readmeFile = 'This Helm chart is generated using cdk8s. Any manual changes to the chart would be discarded once cdk8s app is synthesized again with `--format helm`.';
+
+    fs.outputFileSync(path.join(root, README), readmeFile);
+
+    if (config?.synth?.chartApiVersion === HelmChartApiVersion.V2 && crdsArePresent(config.imports)) {
```
iliapolo: Notice that this *should* have been a bug because you don't consider the default value of `chartApiVersion` here.

```suggestion
    if (apiVersion === HelmChartApiVersion.V2 && crdsArePresent(config?.imports)) {
```

The only reason things work is because you use `fs.outputFileSync` when actually adding the CRDs, which is recursively creating the necessary directories. So we can get rid of this entirely.

src/config.ts
```diff
@@ -12,37 +10,49 @@ export interface ImportSpec {
...
+  PLAIN = 'plain',
+  HELM = 'helm',
+}
+
+export enum HelmChartApiVersion {
+  V1 = 'v1',
+  V2 = 'v2',
+}
+
+export interface Synth {
```
iliapolo:
```suggestion
export interface SynthConfig {
```

src/import/crd.ts
```diff
@@ -125,14 +125,18 @@ export class ImportCustomResourceDefinition extends ImportBase {
   public static async fromSpec(importSpec: ImportSpec): Promise<ImportCustomResourceDefinition> {
     const { source } = importSpec;
     const manifest = await download(source);
-    return new ImportCustomResourceDefinition(safeParseCrds(manifest));
+    return new ImportCustomResourceDefinition(manifest);
   }
 
+  rawManifest: string;
```
iliapolo:
```suggestion
public readonly rawManifest: string;
```

test/synth/synth-stdout.test.ts
```diff
@@ -1,13 +1,19 @@
...
+
+const DEFAULT_APP = 'node index.js';
+const imports: string[] = [];
 
 beforeEach(() => {
   // resetting so that every test can use a different config file,
   // which is read on module load.
   jest.resetModules();
+  // Emptying the imports every time a test is run
+  imports.length = 0;
```
iliapolo: This is highly irregular and is normally a code smell. Lets talk about it online.

iliapolo (changes requested):

src/cli/cmds/synth.ts
vinayak-kukreja: Hmm, wow, sorry I missed this. Curious to learn, how did you catch this? :)

iliapolo: What caught my eye is the use of `config` in a function that accepts inputs that are already derived from the config. In general, there should be a single place in the code that handles every layer, here you had two.

test/synth/synth-stdout.test.ts
```diff
@@ -413,3 +1114,47 @@ function requireSynth() {
...
+    const filePath = path.join(workdir, file);
+
+    if (file === CHART_YAML) {
+      const source = fs.readFileSync(filePath, 'utf-8');
+      const yamlContents = yaml.parse(source);
+
+      expect(yamlContents.name).toMatch(/cdk8s-/);
+      expect(yamlContents.description).toMatch(/Generated chart for cdk8s-/);
+
+      // These values are not stable
```
iliapolo: How so? because its derived from the working directory? If that's the case, lets create a stable directory name inside a temporary directory.

test/synth/synth-stdout.test.ts
```diff
@@ -413,3 +1114,47 @@ function requireSynth() {
...
+      yamlContents.name = omitted;
+      yamlContents.description = omitted;
+
+      map[file] = yaml.stringify(yamlContents);
+      continue;
+    }
+
+    if (file.includes('crds/')) {
+      // crds/encodedFileName are not stable
```
iliapolo: How come? shouldn't this only be derived from the CRD file name?

test/synth/synth-stdout.test.ts
```diff
@@ -317,21 +323,714 @@ describe('validations', () => {
...
+  });
+
+  const importsForChartApiv2 = [
+    'k8s',
+    path.join(__dirname, './__resources__/crds/foo.yaml'),
+    `bar:=${path.join(__dirname, './__resources__/crds/bar.yaml')}`,
+    'github:crossplane/crossplane@0.14.0',
+  ];
+
+  const synthWorksForChartAPIv2WithCrds = async (dir: string) => {
```
iliapolo: Why do we need all these functions? The flow should just be: "synthesize -> expectSnapshot", without any specific comparison functions. No?

test/synth/synth-stdout.test.ts
vinayak-kukreja: The dir now created for test would have consistent name. So this won't be an issue.

test/synth/synth-stdout.test.ts
vinayak-kukreja: This is very strange. The filename does not change when running synth multiple times locally. This is changed when running build on PR: [`be1fe8d` (#1195)](https://github.com/cdk8s-team/cdk8s-cli/pull/1195/commits/be1fe8d0a636474fad18c00ae21b048b5bd91ac5)

Taking a look.

vinayak-kukreja: So this was happening because the hash was being calculated for the file with its path. This changes when being run on PR.

I have changed it to just hash and encode for file name if it is not a url.

test/util.test.ts
```diff
@@ -46,4 +46,45 @@ describe('findManifests', () => {
     //Bad directory name (non-existent) should yield an empty array
     expect(noFilesEither).toEqual([]);
   });
+});
+
+test('derive file name from url', () => {
+  const devUrl = 'github:crossplane/crossplane@0.14.0';
+  const localFile = './foo/bar/baz/fooz.yaml';
```
iliapolo: Add more scenarios:

- local file that ends with `.yml` (its also a common YAML extension)
- regular URL (not `github:`)

iliapolo (changes requested):

iliapolo (approved):

PR #1202, opened 2023-08-08
feat: helm import

vinayak-kukreja:
Resolves https://github.com/cdk8s-team/cdk8s-team-internal/issues/54


src/import/dispatch.ts
```diff
@@ -30,6 +31,12 @@ async function matchImporter(importSpec: ImportSpec, argv: any): Promise<ImportB
     return new ImportKubernetesApi(k8s);
   }
 
+  const helm = importSpec.source.split(':')[0];
```
iliapolo:
```suggestion
  const prefix = importSpec.source.split(':')[0];
```

src/import/dispatch.ts
```diff
@@ -30,6 +31,12 @@ async function matchImporter(importSpec: ImportSpec, argv: any): Promise<ImportB
...
+
+  if (helm === 'helm') {
+    return new ImportHelm(importSpec);
```
iliapolo: For consistency with other importers:

 ```suggestion
    return ImportHelmChart.fromSpec(importSpec);
```

File: src/import/helm.ts [hunk 3]
```diff
@@ -0,0 +1,173 @@
...
+const CHART_YAML = 'chart.yaml';
+
+// TODO: Validate if chart version follows SemVer
+export class ImportHelm extends ImportBase {
+  chartName: string;
+  chartUrl: string;
+  chartVersion: string;
+  chartSchemaPath: string | undefined;
+  chartDependencies: string[] = [];
+  tmpDir: string;
```
iliapolo: These should all be `readonly`.

src/import/helm.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    const [chartUrl, chartName, chartVersion] = getHelmChartDetails(importSpec.source);
+
+    this.chartName = chartName;
+    this.chartUrl = chartUrl;
+    this.chartVersion = chartVersion;
+    this.tmpDir = pullHelmRepo(chartUrl, chartName, chartVersion);
+
+    const chartYamlFilePath = path.join(this.tmpDir, this.chartName, CHART_YAML);
+    const contents = Yaml.load(chartYamlFilePath)[0];
```
iliapolo: Can you explain the `[0]`? Should we add some protective code here?

src/import/helm.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    const potentialSchemaPath = path.join(this.tmpDir, this.chartName, CHART_SCHEMA);
+
+    if (fs.existsSync(potentialSchemaPath)) {
+      this.chartSchemaPath = potentialSchemaPath;
+    } else {
+      this.chartSchemaPath = undefined;
+    }
+
+    console.log(`CHART SCHEMA PATH: ${this.chartSchemaPath}`);
```
iliapolo: Can we cleanup the temp directory here? I believe all the necessary information has already been extracted at this point?

src/import/codegen.ts
```diff
@@ -213,3 +213,130 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
...
+        code.line('readonly values?: { [key: string]: any };');
+      } else {
+        code.line(`readonly values?: ${valuesInterface};`);
+      }
+
+      code.closeBlock();
+    }
+
+    function emitConstruct() {
+      code.openBlock(`export class ${chartName} extends Helm`);
```
iliapolo: I don't think this is going to fly with strongly typed languages because it will require we have a constructor matching the `Helm` class one, which we absolutely don't want. I think we should just extend `Construct` here like normal, not sure there is value in extending `Helm` since we are replicating the helm properties anyway. (e.g namespace, releaseName, ...).

src/import/codegen.ts
```diff
@@ -213,3 +213,130 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
...
+    if (schema !== undefined) {
+      // Interface for schema generated props
+      let schemaGenValuesInterface: string = 'SchemaGeneratedValues';
+      schemaGenValuesInterface = typegen.emitType(schemaGenValuesInterface, schema, def.fqn);
+
+      // Creating values interface
+      emitValuesInterface();
+
+      function emitValuesInterface() {
+        code.openBlock(`export interface ${valuesInterface} extends ${schemaGenValuesInterface}`);
```
iliapolo: This complicates the code for no good reason I feel. I understand the distinction between props that come from the schema, and props that we populate - but in practice it doesn't really matter, especially for the user who might want to look at the code. Lets simplify but just putting everything in a single interface.

src/import/codegen.ts
```diff
@@ -213,3 +213,130 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
...
+  const repoUrl = def.chartUrl;
+  const chartVersion = def.chartVersion;
+
+  // Create custom type
+  typegen.emitCustomType(chartName, code => {
+
+    const valuesInterface = `${chartName}ValuesProps`;
+    if (schema !== undefined) {
+      // Interface for schema generated props
+      let schemaGenValuesInterface: string = 'SchemaGeneratedValues';
```
iliapolo: Can we just call this `Values`?

src/import/codegen.ts
```diff
@@ -213,3 +213,130 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
...
+    function emitPropsInterface() {
+      code.openBlock(`export interface ${chartName}Props`);
+
+      code.line('readonly namespace?: string;');
+      code.line('readonly releaseName?: string;');
+      code.line('readonly helmExecutable?: string;');
+      code.line('readonly helmFlags?: string[];');
+
+      if (schema === undefined) {
+        code.line('readonly values?: { [key: string]: any };');
```
iliapolo, line 338: Would be really nice to check if there are any required properties in the schema, and then mark `values` as required - which would then need to be propagated to the entire `props` as required.

iliapolo (changes requested):

src/import/helm.ts
vinayak-kukreja: I am just getting the chart.yaml from the helm chart. You are correct, I will add checks to this.

src/import/helm.ts
vinayak-kukreja: It is being cleaned up after we use the schema: https://github.com/cdk8s-team/cdk8s-cli/pull/1202/files#diff-609ab80fc6736ca9bfbc54b4379e2bb356360d6f479d9e23a193d5a1299bb727R86

src/import/codegen.ts
vinayak-kukreja: Hmm, makes sense. Will update this.

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+    const { source } = importSpec;
+    return new ImportHelm(source);
+  }
+
+  public readonly chartName: string;
+  public readonly chartUrl: string;
+  public readonly chartVersion: string;
+  public readonly chartSchemaPath: string | undefined;
+  public readonly chartDependencies: string[] = [];
+  public readonly tmpDir: string;
```
iliapolo: Do these need to be `public`? Looks like they can be `private`

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
+import { spawnSync } from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { Yaml } from 'cdk8s';
+import { CodeMaker } from 'codemaker';
+// we just need the types from json-schema
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { JSONSchema4 } from 'json-schema';
```
iliapolo:
```suggestion
import type { JSONSchema4 } from 'json-schema';
```

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+      }
+    }
+
+    const potentialSchemaPath = path.join(this.tmpDir, this.chartName, CHART_SCHEMA);
+
+    if (fs.existsSync(potentialSchemaPath)) {
+      this.chartSchemaPath = potentialSchemaPath;
+    } else {
+      this.chartSchemaPath = undefined;
+    }
```
iliapolo:
```suggestion
    this.chartSchemaPath = fs.existsSync(potentialSchemaPath) ? potentialSchemaPath : undefined
```

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    this.chartName = chartName;
+    this.chartUrl = chartUrl;
+    this.chartVersion = chartVersion;
+    this.tmpDir = pullHelmRepo(chartUrl, chartName, chartVersion);
+
+    const chartYamlFilePath = path.join(this.tmpDir, this.chartName, CHART_YAML);
+    const contents = Yaml.load(chartYamlFilePath);
+
+    if (contents && contents.length === 1 && contents[0].dependencies) {
```
iliapolo: `contents` is never undefined.

```suggestion
    if (contents.length === 1 && contents[0].dependencies) {
```

Also I don't understand the `[0]`...what happens if `contents.length > 1`? what does this mean? can you paste an example `Chart.yaml` with dependencies?

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+      for (const dependency of contents[0].dependencies) {
+        this.chartDependencies.push(dependency.name);
+      }
+    }
+
+    const potentialSchemaPath = path.join(this.tmpDir, this.chartName, CHART_SCHEMA);
+
+    if (fs.existsSync(potentialSchemaPath)) {
```
iliapolo: Looks like `this.tmpDir` can be cleaned up right after this right?

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+  protected async generateTypeScript(code: CodeMaker) {
+
+    emitHelmHeader(code);
+
+    let schema: JSONSchema4 | undefined;
+    if (this.chartSchemaPath !== undefined) {
+      schema = JSON.parse(fs.readFileSync(this.chartSchemaPath, 'utf-8'));
+    } else {
+      schema = undefined;
+    }
```
iliapolo: Use a trinary expression and move to the constructor so that this method has no dependencies on the file system

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+ * @param url
+ */
+function validateHelmUrl(url: string): RegExpExecArray {
+  const helmRegex = /^helm:([A-Za-z0-9_.-:\-]+)\/([A-Za-z0-9_.-:\-]+)\@([0-9]+)\.([0-9]+)\.([A-Za-z0-9-+]+)$/;
+  const match = helmRegex.exec(url);
+
+  if (match) {
+    return match;
+  } else {
+    throw Error(`There was an error processing the helm chart url you passed in: ${url}. Make sure it matches the format of 'helm:<repo-url>/<chart-name>@<chart-version>'.`);
```
iliapolo:
```suggestion
    throw Error(`Invalid helm URL: ${url}. Must match the format: 'helm:<repo-url>/<chart-name>@<chart-version>'.`);
```

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    cleanup(this.tmpDir);
+  }
+}
+
+/**
+ * Validating if a helm chart url is in an expected format
+ * @param url
+ */
+function validateHelmUrl(url: string): RegExpExecArray {
```
iliapolo: No need to a function here, it creates an awkward return value...just inline this logic in `getHelmChartDetails`

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+    throw Error(`There was an error processing the helm chart url you passed in: ${url}. Make sure it matches the format of 'helm:<repo-url>/<chart-name>@<chart-version>'.`);
+  }
+}
+
+/**
+ * Gets information about the helm chart from the helm url
+ * @param url
+ * @returns chartUrl, chartName and chartVersion
+ */
+function getHelmChartDetails(url: string) {
```
iliapolo: Make sure to use the right semantics for function names, it makes it easier to understand. (`get` is for getters, `fetch` is for http calls, `extract` is for regex and such...)

```suggestion
function extractHelmChartDetails(url: string) {
```

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+  const chartUrl = helmDetails[1];
+  const chartName = helmDetails[2];
+  const major = helmDetails[3];
+  const minor = helmDetails[4];
+  const patch = helmDetails[5];
+
+  const chartVersion = `${major}.${minor}.${patch}`;
+
+  if (!semver.valid(chartVersion)) {
+    throw new Error(`The value specified in '${url}' for chart version: '${chartVersion}' does not follow SemVer-2(https://semver.org/).`);
```
iliapolo:
```suggestion
    throw new Error(`Invalid chart version (${chartVersion}) in URL: ${url}. Must follow SemVer-2 (see https://semver.org/).`);
```

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+  const command = 'helm';
+
+  const helm = spawnSync(command, args, {
+    maxBuffer: MAX_HELM_BUFFER,
+  });
+
+  if (helm.error) {
+    const err = helm.error.message;
+    if (err.includes('ENOENT')) {
+      throw new Error(`unable to execute '${command}' to pull the Helm chart. Is it installed on your system?`);
```
iliapolo: Nice! 👍

iliapolo:
```suggestion
      throw new Error(`Unable to execute '${command}' to pull the Helm chart. Is helm installed on your system?`);
```

src/import/helm.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    throw new Error(`error while pulling a helm chart: ${err}`);
```
iliapolo:
```suggestion
    throw new Error(`Failed pulling helm chart from URL (${chartUrl}): ${err}`);
```

src/import/helm.ts
vinayak-kukreja: Ok, now reading schema in this function and deleting temp dir. Let me know if that works.

src/import/helm.ts
vinayak-kukreja: Ohh, the `contents[0]` is because `Yaml.load` returns a list,
```
(method) Yaml.load(urlOrFile: string): any[]

Downloads a set of YAML documents (k8s manifest for example) from a URL or a file and returns them as javascript objects.

Empty documents are filtered out.

@param urlOrFile — a URL of a file path to load from

@returns — an array of objects, each represents a document inside the YAML
```

You are correct, it would never be undefined. Will update it. 

`contents.length > 1` should not happen since helm chart should just have one `Chart.yaml`. I can switch to reading and parsing the Yaml file instead of Yaml.load. Let me know if that would be more readable.

vinayak-kukreja: Here is a Chart.yaml with dependencies: https://github.com/bitnami/charts/blob/main/bitnami/airflow/Chart.yaml#L22-L35

iliapolo (approved):

PR #1315, opened 2023-08-30
fix(import): incorrect and unnecessary properties recorder to `cdk8s.yaml` on `cdk8s import`

### LINKED ISSUE #1442: cdk8s import behavior changed?
### Description of the bug:

Looks like `cdk8s import` command is recently changed.

### Reproduction Steps:

`cdk8s.yaml` content:

```yml
language: typescript
imports:
  - 1.24:=k8s@1.24.0
  - 1.25:=k8s@1.25.0
  - 1.26:=k8s@1.26.0
  - ....
```

When I run `npx cdk8s import --output packages/my-lib/imports`, `cdk8s.yaml` is updated to

```yml
output: dist # ??
pluginsDirectory: /Users/MYUSERNAME/.cdk8s/plugins # ??
language: typescript
imports:
  - 1.24:=k8s@1.24.0
  - 1.25:=k8s@1.25.0
  - 1.26:=k8s@1.26.0
  - k8s@1.24.0 # ??
  - k8s@1.25.0 # ??
  - k8s@1.26.0 # ??
  - ....
```

### Environment:

  - **Framework Version**:
  - **OS**:
```
❯ npx cdk8s --version
2.50.0
```

### Other:

---

This is :bug: Bug Report

iliapolo:
When an import is specified with an alias, i.e `cdk8s import 1.24:=k8s@1.24`, the alias is being ignored when saving it to the configuration file. This creates duplicate import entries, one with the alias and one without. 

In addition, default configuration values that were only meant for in-process operation are written to the config file on import - this should not happen.

Also added a `--no-save` flag that skips writing the import to the file. This can be helpful if users have comments in `cdk8s.yaml`, which cannot be preserved upon rewrite. 

Fixes https://github.com/cdk8s-team/cdk8s/issues/1442

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -46,7 +46,7 @@ export function readConfigSync(): Config {
 }
 
 export async function addImportToConfig(source: string) {
-  let curConfig = readConfigSync();
+  let curConfig = yaml.parse(fs.readFileSync(CONFIG_FILE, 'utf-8'));
 
   const curImports = curConfig.imports ?? [];
   if (!curImports.includes(source)) {
--- a/src/import/dispatch.ts
+++ b/src/import/dispatch.ts
@@ -21,7 +21,7 @@ export async function importDispatch(imports: ImportSpec[], argv: any, options: 
       ...options,
     });
 
-    if (options.save ?? false) {
+    if (options.save ?? true) {
       const spec = importSpec.moduleNamePrefix ? `${importSpec.moduleNamePrefix}${PREFIX_DELIM}${importSpec.source}` : importSpec.source;
       await addImportToConfig(spec);
     }
--- a/src/import/dispatch.ts
+++ b/src/import/dispatch.ts
@@ -4,6 +4,8 @@ import { matchCrdsDevUrl } from './crds-dev';
 import { ImportKubernetesApi } from './k8s';
 import { ImportSpec, addImportToConfig } from '../config';
 
+export const PREFIX_DELIM = ':=';
+
 export async function importDispatch(imports: ImportSpec[], argv: any, options: ImportOptions) {
   for (const importSpec of imports) {
     const importer = await matchImporter(importSpec, argv);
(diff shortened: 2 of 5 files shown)


src/config.ts
```diff
@@ -46,7 +46,7 @@ export function readConfigSync(): Config {
 }
 
 export async function addImportToConfig(source: string) {
-  let curConfig = readConfigSync();
```
iliapolo, line 49: This what caused the defaults to be written. Now we just read the current file on disk, without the defaults, and add upon it.

src/import/dispatch.ts
```diff
@@ -18,7 +20,11 @@ export async function importDispatch(imports: ImportSpec[], argv: any, options:
       moduleNamePrefix: importSpec.moduleNamePrefix,
       ...options,
     });
-    await addImportToConfig(importSpec.source);
+
+    if (options.save ?? false) {
+      const spec = importSpec.moduleNamePrefix ? `${importSpec.moduleNamePrefix}${PREFIX_DELIM}${importSpec.source}` : importSpec.source;
```
iliapolo, line 25: This is where the alias should have been considered and now is.

test/import/import-crd.test.ts
```diff
@@ -597,16 +597,36 @@ describe('cdk8s.yaml file', () => {
...
 
   test('does not update with CRD that is imported twice', async () => {
     await importDispatch([jenkinsCRD], {}, importOptions);
     await importDispatch([jenkinsCRD], {}, importOptions);
 
-    const config = readConfigSync();
-    expect(config.imports?.length == 2).toBeTruthy();
+    const config = yaml.parse(fs.readFileSync('cdk8s.yaml', 'utf-8'));
+    expect(config).toMatchSnapshot();
   });
```
sumupitchayan, line 610: Should we also test that comments in the cdk8s.yaml file are not deleted with the `--no-save` flag?

iliapolo: We already have a test that makes sure the file isn't touched with `--no-save`

sumupitchayan (approved):

PR #1552, opened 2023-10-17
fix(import):  `flattenAdditionalValues` does not handle arrays properly

### LINKED ISSUE #1538: flattenAddtionalValues generated during helm import does not handle arrays correctly
### Description of the bug:
After importing a helm chart, and adding array values and running `synth` I found that the yaml that is generated is incorrect, it creates a map instead of a list. In this instance, I am using the minio operator chart, but I've tried with a couple of others, examples below.

### Reproduction Steps:

cdk8s.yml
```yaml
language: typescript
app: npx ts-node main.ts
imports:
  - k8s
  - minio:=helm:https://operator.min.io/operator@5.0.9
  - helm:https://grafana.github.io/helm-charts/loki@5.27.0
```

MINio chart
```typescript
export class MyChart extends Chart {
  constructor(scope: Construct, id: string, props: ChartProps = {}) {
    super(scope, id, props);

    new MinioOperator(this, "minio-operator", {
      values: {
        operator: {
          replicaCount: 1,
          volumes: [
            {
              mountPath: "/tmp/certs",
              name: "tls-minio",
            },
          ],
          volumeMounts: [
            {
              name: "tls-minio",
              projected: {
                defaultMode: 420,
                sources: [
                  {
                    secret: {
                      name: "tls-minio",
                      items: [
                        {
                          key: "tls.crt",
                          path: "CAs/custom-public.crt",
                        },
                      ],
                    },
                  },
                  {
                    secret: {
                      name: "tls-minio",
                      items: [
                        {
                          key: "tls.crt",
                          path: "CAs/custom-public.crt",
                        },
                      ],
                    },
                  },
                ],
              },
            },
          ],
        },
      },
    });
  }
}
```

synth output:
```yaml
---
volumeMounts:
  "0":
    name: tls-minio
    projected:
      defaultMode: 420
      sources:
        "0":
          secret:
            items:
              "0":
                key: tls.crt
                path: CAs/custom-public.crt
            name: tls-minio
        "1":
          secret:
            items:
              "0":
                key: tls.crt
                path: CAs/custom-public.crt
            name: tls-minio
---
volumes:
  "0":
    mountPath: /tmp/certs
    name: tls-minio
```

The same happens with Loki, but this time I get an error message in console instead of a synthed chart

Chart
```yaml
export class MyChart extends Chart {
  constructor(scope: Construct, id: string, props: ChartProps = {}) {
    super(scope, id, props);

    new Loki(this, "loki", {
      releaseName: "loki",
      values: {
        loki: {
          auth_enabled: false,
          replication_enabled: false,
        },
        common_config: {
          replication_factor: 1,
        },
        write: {
          replicas: 1,
        },
        read: {
          replicas: 1,
        },
        backend: {
          replicas: 1,
          extraVolumes: [
            {
              mountPath: "/tmp/certs",
              name: "tls-minio",
            },
          ],
        },
        test: {
          enabled: false,
        },
      },
    });
  }
}
```

### Error Log:
```
...\cdk8s\src\helm.ts:152
    throw new Error(helm.stderr.toString());
          ^
Error: coalesce.go:223: warning: destination for loki.backend.extraVolumes is a table. Ignoring non-table value ([])
coalesce.go:223: warning: destination for loki.backend.extraVolumes is a table. Ignoring non-table value ([])
Error: YAML parse error on loki/templates/backend/statefulset-backend.yaml: error converting YAML to JSON: yaml: line 137: did not find expected '-' indicator
```

### Environment:

  - **Framework Version**: 2.138.0
  - **OS**: Windows 22H2 Build 22621.2428 

### Other:
After reviewing the `flattenAdditionaValues` function in the generated import it looks as though arrays are not explicitly handled, maybe something like this inside the for loop:

```typescript
for (let prop in props) {
    if (Array.isArray(props[prop])) {
      output[prop] = props[prop].map(item => {
        if (typeof item === 'object') {
          return this.flattenAdditionalValues(item);
        }
        return item;
      });
    } 
    else if (typeof(props[prop]) === 'object' && prop !== 'additionalValues') {
      output[prop] = this.flattenAdditionalValues(props[prop]);
    } 
    else {
      output[prop] = props[prop];
    }
  }
```
---

This is :bug: Bug Report

vinayak-kukreja:
Currently, `flattenAdditionalValues` is not handling arrays correctly since these are also handled similar to `objects`. Adding functionality to support this. 
* Also updating `Helm` construct id to unique `ids`. 
* And, adding `:=helm:` to `isHelmImport` function.

Fixes https://github.com/cdk8s-team/cdk8s-cli/issues/1538

### DIFF
--- a/src/import/codegen.ts
+++ b/src/import/codegen.ts
@@ -384,14 +384,21 @@ export function generateHelmConstruct(typegen: TypeGenerator, def: HelmObjectDef
       code.close('};');
 
       code.line();
-      code.line('new Helm(scope, \'Helm\', finalProps)');
+      code.line('new Helm(scope, `Helm${id`f}`, finalProps)');
       code.closeBlock();
     }
 
     function emitAdditionalValuesFlattenFunc() {
       code.openBlock('private flattenAdditionalValues(props: { [key: string]: any }): { [key: string]: any }');
       code.open('for (let prop in props) {');
-      code.open('if (typeof(props[prop]) === \'object\' && prop !== \'additionalValues\') {');
+      code.open('if (Array.isArray(props[prop])) {');
+      code.open('props[prop].map((item: any) => {');
+      code.open('if (typeof(item) === \'object\' && prop !== \'additionalValues\') {');
+      code.line('return this.flattenAdditionalValues(item);');
+      code.close('}');
+      code.line('return item;');
+      code.close('});');
+      code.open('} else if (typeof(props[prop]) === \'object\' && prop !== \'additionalValues\') {');
       code.line('props[prop] = this.flattenAdditionalValues(props[prop]);');
       code.close('}');
       code.close('}');
--- a/src/import/codegen.ts
+++ b/src/import/codegen.ts
@@ -360,7 +360,7 @@ export function generateHelmConstruct(typegen: TypeGenerator, def: HelmObjectDef
       const propsDefinition = schema && hasRequiredProps(schema) ? `${chartName}Props` : `${chartName}Props = {}`;
 
       code.openBlock(`public constructor(scope: Construct, id: string, props: ${propsDefinition})`);
-      code.line('super(scope, id)');
+      code.line('super(scope, id);');
 
       code.line('let updatedProps = {};');
       code.line();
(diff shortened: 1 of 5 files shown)


src/import/codegen.ts
```diff
@@ -384,14 +384,21 @@ export function generateHelmConstruct(typegen: TypeGenerator, def: HelmObjectDef
       code.close('};');
 
       code.line();
-      code.line('new Helm(scope, \'Helm\', finalProps)');
+      code.line('new Helm(scope, `Helm${id}`, finalProps)');
```
iliapolo: Whoopsy, we missed passing the correct scope. 

```suggestion
      code.line('new Helm(this, 'Helm', finalProps)');
```

iliapolo (changes requested):

iliapolo (approved):

PR #1629, opened 2023-11-09
Adding deno flavor of typescript as an app type

tskinn:
Apologies I haven't created a ticket for this to discuss but it didn't take long to get something working with which I could start a conversation.

Is this something the project would consider? I find the idea of having no separate dependencies(node_modules) management and no compile step creating extra files to be very desirable. 

The code changes are not ready for serious review but I wanted to open this PR as I had the code handy and wanted to start a conversation. What needs to be done for this to progress if it is an acceptable change?

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -172,6 +172,34 @@
         }
       ]
     },
+    "integ:init:deno-app-npm": {
+      "name": "integ:init:deno-app-npm",
+      "steps": [
+        {
+          "exec": "yarn run compile"
+        },
+        {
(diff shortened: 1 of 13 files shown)


iliapolo (commented):
@tskinn appreciate the work you've done here, but we are not going to include this in our base templates:

- This is not a new language, but rather a different runtime.
- We don't currently have official support for deno and are not planning on it.

Users are free to adjust the typescript project created with `cdk8s init typescript-app` to use the deno runtime. 

Thank you.

PR #1634, opened 2023-11-10
fix(util): correct logging for synth

iceglober:
FYI - This is my first OSS contribution. I'm open to any feedback or input!

# Summary of Changes
The console logs generated during `cdk8s synth` gave the impression that synthesis had failed when it actually succeeded, leading to unnecessary confusion. This change updates that log to a warning if the `outdir` doesn't exist, and also adds a "success" log if generated manifests are found. This isn't actually a problem because [`synth` in cdk8s-core](https://github.com/cdk8s-team/cdk8s-core/blob/640a136333e0f1303a3979d3c6e8ba3e937b2a92/src/app.ts#L184) always attempts to create the necessary directory.

# Current State
Without providing a CLI option for `outdir` and without specifying an `outdir` property in my config file, I defined my App like so:
```typescript
const app = new App({
    outdir: 'out',
    yamlOutputType: YamlOutputType.FOLDER_PER_CHART_FILE_PER_RESOURCE,
});
new MongoDbChart(app, 'mongoDb2');
app.synth();
```
where `outdir` is defined in my IaC directly. I wanted to make a clear differentiation between TS output ("dist") and `cdk8s` output (I chose "out").

With this current state, here's what the console output looks like:
```bash
iceglobe@iceglobe-MS-7C79 ~/projects/kayn/k8s/cdk8s
 % cdk8s synth
Synthesizing application
ERROR: synthesis failed, app expected to create "dist"
```
This "ERROR" message gives the false impression that manifest synthesis failed, when in reality, because I had specified a different `outdir` in my TS file, everything worked just fine.

# Proposed Change
My changes make it so that the console output reflects the following:
```bash
iceglobe@iceglobe-MS-7C79 ~/projects/kayn/k8s/cdk8s
 % cdk8s synth
WARNING: specified outdir does not exist, synth command could fail
Synthesizing application
  - dist/mongodb2-c84b3744/Deployment.mongodb.k8s.yaml
  - dist/mongodb2-c84b3744/Secret.mongodb-secret.k8s.yaml
  - dist/mongodb2-c84b3744/Service.mongodb-service.k8s.yaml
Manifests synthesized!
```
This accomplishes the following:
- Gives warning that synth might fail depending on the shell command that is run
- Allows intended logging of generated manifests to be displayed (this was previously inhibited because the `pathExists` check resulted in an exit)
- Gives an affirmative message that manifests were indeed generated as expected ("Manifests synthesized!")


src/util.ts
```diff
@@ -44,9 +44,14 @@ export async function mkdtemp(closure: (dir: string) => Promise<void>) {
 }
 
 export async function synthApp(command: string, outdir: string, stdout: boolean, metadata: boolean): Promise<SynthesizedApp> {
+  if (!await fs.pathExists(outdir)) {
```
iliapolo, line 47: Putting this here means that this warning will **always** be printed because the outdir is expected to be created further down (in the `await shell` call).

src/util.ts
```diff
@@ -76,6 +76,8 @@ export async function synthApp(command: string, outdir: string, stdout: boolean,
 
   if (!found) {
     console.error('No manifests synthesized');
+  } else {
+    console.log('Manifests synthesized!')
```
iliapolo, line 80: I don't think this necessary though. If we found manifests, then we [print their path](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/util.ts#L71) already, indicating a successful synthesis.

src/util.ts
```diff
@@ -58,11 +63,6 @@ export async function synthApp(command: string, outdir: string, stdout: boolean,
     },
   });
 
-  if (!await fs.pathExists(outdir)) {
-    console.error(`ERROR: synthesis failed, app expected to create "${outdir}"`);
-    process.exit(1);
```
iliapolo, line 63: I understand where your coming from with having `synth` not fail in this case (because manifests were actually created). However, from the CLIs perspective, synthesis is not only creating manifests, there are additional CLI features (like validations) that rely on the fact this directory exists, and so from its POV, synthesis is not successful. This does unfortunately mean that, if you define `outdir` in your app, you must also change the `output` property in the `cdk8s.yaml` configuration file. 

Note that if you do it the other way around, i.e set `output: out` in `cdk8s.yaml`, and leave the `App` instance with its defaults, everything will work as expected.

iliapolo (commented):
@iceglober Appreciate the work here, but we can't accept this change because it inherently changes the behavior of the CLI. Also as I explain in the comments, the current behavior is the desired one, albeit somewhat confusing. 

Thanks!

iliapolo:
Apologies for the long delay in reviewing this.

PR #1666, opened 2023-11-21
fix: respect hyphens in helm chart names

rassie:
From `json2jsii` 0.4.0 onwards, type name sanitization [works correctly with kebab case](https://github.com/cdklabs/json2jsii/pull/1244), i.e. it no longer ignores hyphens, so that names like `ingress-nginx`, which previously have been previously converted to `Ingressnginx`, now get converted to `IngressNginx`, which is the expected behaviour for kebab case to pascal case conversion. However, `cdk8s-cli`'s helm import has been additionally sanitizing hyphens in chart names, even though hyphens are valid characters in chart names, so that the change in `json2jsii` had no effect. This patch simplifies the helm importer's sanitization routine to not remove hyphens anymore, resulting in more correct class names.

This will be a breaking change for people using Helm charts with a hyphen in the name.


rassie:
cc @vinayak-kukreja, since you are the author of the original patch

vinayak-kukreja:
Hey @rassie, thank you for opening a contribution with us and apologies for the delay. 

I am not sure if we can accept a breaking change. Probably would need to be behind some kind of feature flag. Let me consult the team and get back to you.

shinebayar-g:
https://github.com/cdk8s-team/cdk8s-cli/pull/1737/files was a breaking change to the same effect.
https://github.com/cdk8s-team/cdk8s-cli/compare/v2.198.4...v2.198.5

iliapolo:
@rassie Since its a breaking change, we need to make this optional...I don't mind introducing a poor mans feature toggle in the form of an environment variable, something like: `CDK8S_HELM_IMPORT_RESPECT_HYPHENS`.

@shinebayar-g 

> https://github.com/cdk8s-team/cdk8s-cli/pull/1737/files was a breaking change to the same effect.
https://github.com/cdk8s-team/cdk8s-cli/compare/v2.198.4...v2.198.5

True, and a good call out. This has surfaced an automation problem with how we bump and approve dependency upgrades for 0.x versions. We need to address this. But I wouldn't want to introduce yet another breaking change because we once let one slip.

rassie:
The feature toggle looks quite ugly, if I'm being completely honest. This PR has been open for quite a while, from my POV it'd probably be fine if it just became part of an eventually released major version.

The question is though whether you'd actually _want_ to release a new major version, judging by how 1.x is still a thing three years after initial 2.x release, meaning you'd probably have three major versions around (along with 5-6 `cdk8s+` versions). However, if you are open to it, a new major might be a better way.

rassie:
And just as I point it out, v1 seems to be on its way out (https://github.com/cdk8s-team/cdk8s-cli/pull/2710). Sorry, haven't seen it while writing the comment.

iliapolo:
@rassie Ok, lets keep it around for a future 3.x version line. If we see more community demand for this we might reconsider the "feature toggle", which I agree is quite ugly. I can tell you that we already have a few other issues in the pipeline that require a major version bump, so this just gives us more incentive to roll it out. But I can't attest to when that will be.

iliapolo:
Converting to draft until we decide how and when to include this.

PR #1715, opened 2023-12-04
chore: upgrade configuration

vinayak-kukreja:
Currently the workflow is broken and is not able to upgrade configuration. With this upgrade the workflow file would fix and github workflow would be able to upgrade this on its own.


.github/workflows/upgrade-compiler-dependencies-1.x.yml
```diff
@@ -4,7 +4,7 @@ name: upgrade-compiler-dependencies-1.x
 on:
   workflow_dispatch: {}
   schedule:
-    - cron: 0 0 * * *
+    - cron: 0 12 * * *
```
sumupitchayan, line 7: @vinayak-kukreja is there a reason some of these schedules are changed?

vinayak-kukreja: Yeah, so unfortunately our workflow has been broken since 4-5 months. This was added in this commit: https://github.com/cdk8s-team/cdk8s-projen-common/commit/628d3c07b35bd3875d0a5aef5e5c89372b56ac16 but never made it here.

sumupitchayan (approved):

.github/workflows/upgrade-compiler-dependencies-1.x.yml
vinayak-kukreja: Our workflow: https://github.com/cdk8s-team/cdk8s-cli/actions/workflows/upgrade-configuration-2.x.yml

PR #1738, opened 2023-12-08
fix: code generating enum values containing !, =, ~ are stripped

### LINKED ISSUE #229: non alpha-numeric enum values cannot be imported
For example: https://github.com/prometheus-operator/kube-prometheus/blob/373e2b415bd0ddbcb049dea54db80c2a78b76945/manifests/setup/0alertmanagerConfigCustomResourceDefinition.yaml#L67-L71

This results in:

```ts
/**
 * Match operation available with AlertManager >= v0.22.0 and takes precedence over Regex (deprecated) if non-empty.
 *
 * @schema AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType
 */
export enum AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType {
  /** __stripped_by_cdk8s__ */
  STRIPPED_BY_CDK8S = '__stripped_by_cdk8s__',
  /** __stripped_by_cdk8s__ */
  STRIPPED_BY_CDK8S = '__stripped_by_cdk8s__',
  /** __stripped_by_cdk8s__ */
  STRIPPED_BY_CDK8S = '__stripped_by_cdk8s__',
  /** __stripped_by_cdk8s__ */
  STRIPPED_BY_CDK8S = '__stripped_by_cdk8s__',
}
```

The reason is that these characters are not allowed as enum values. 
However, even if we do allow them, it still won't work:

```ts
/**
 * Match operation available with AlertManager >= v0.22.0 and takes precedence over Regex (deprecated) if non-empty.
 *
 * @schema AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType
 */
export enum AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType {
  /** != */
  VALUE_ = '!=',
  /** = */
  VALUE_ = '=',
  /** =~ */
  VALUE_ = '=~',
  /** !~ */
  VALUE_ = '!~',
}
```

This happens in [json2jsii](https://github.com/cdklabs/json2jsii/blob/main/src/type-generator.ts#L464), because we cannot deduce a member name. 

So solving this issue would require:

1. Make `json2jsii` support these enum values and append the value index to the member name in such cases.
2. Allow these characters in the cdk8s cli.

vinayak-kukreja:
This has been fixed in json2jsii. Allowlisting characters `!, =, ~`. 

Fixes https://github.com/cdk8s-team/cdk8s-cli/issues/229

### DIFF
--- a/test/import/fixtures/kube-prometheus.yaml
+++ b/test/import/fixtures/kube-prometheus.yaml
@@ -1,0 +1,4296 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.13.0
+    operator.prometheus.io/version: 0.70.0
+  name: alertmanagerconfigs.monitoring.coreos.com
+spec:
+  group: monitoring.coreos.com
+  names:
+    categories:
+    - prometheus-operator
+    kind: AlertmanagerConfig
+    listKind: AlertmanagerConfigList
+    plural: alertmanagerconfigs
+    shortNames:
+    - amcfg
+    singular: alertmanagerconfig
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: AlertmanagerConfig configures the Prometheus Alertmanager, specifying how alerts should be grouped, inhibited and notified to external systems.
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: AlertmanagerConfigSpec is a specification of the desired behavior of the Alertmanager configuration. By definition, the Alertmanager configuration only applies to alerts for which the `namespace` label is equal to the namespace of the AlertmanagerConfig resource.
+            properties:
+              inhibitRules:
+                description: List of inhibition rules. The rules will only apply to alerts matching the resource's namespace.
+                items:
+                  description: InhibitRule defines an inhibition rule that allows to mute alerts when other alerts are already firing. See https://prometheus.io/docs/alerting/latest/configuration/#inhibit_rule
+                  properties:
+                    equal:
+                      description: Labels that must have an equal value in the source and target alert for the inhibition to take effect.
+                      items:
+                        type: string
+                      type: array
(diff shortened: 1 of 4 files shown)


test/import/fixtures/kube-prometheus.yaml
```diff
@@ -0,0 +1,4296 @@
...
+                      items:
+                        description: Matcher defines how to match on alert's labels.
+                        properties:
+                          matchType:
+                            description: Match operation available with AlertManager >= v0.22.0 and takes precedence over Regex (deprecated) if non-empty.
+                            enum:
+                            - '!='
+                            - =
+                            - =~
+                            - '!~'
```
vinayak-kukreja, line 58: These are now working. Unfortunately the diff is huge so it does not show up in the web ui. But, you can take this locally and verify it. 

This is what it looks now,

```
/**
 * Match operation available with AlertManager >= v0.22.0 and takes precedence over Regex (deprecated) if non-empty.
 *
 * @schema AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType
 */
export enum AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType {
  /** != */
  VALUE_NOT_EQUALS_TO = \\"!=\\",
  /** = */
  VALUE_ = \\"=\\",
  /** =~ */
  VALUE_EQUAL_TILDE = \\"=~\\",
  /** !~ */
  VALUE_NEGATION_TILDE = \\"!~\\",
}
```

vinayak-kukreja: `  VALUE_ = \\"=\\",`

Fixing this in Json2Jsii now

paulhcsun: What is the purpose of this file?

vinayak-kukreja: So this is a Custom Resource Definition that is failing in the issue mentioned. We can add these to the tests and they create a snapshot that we can verify if the generated code is as expected. 

This CRD was failing code generation due to symbols that were not handled correctly by Json2Jsii. So I fixed that first and then allowlisted these in cdk8s. 

Since `=` was missed, the code generation just gave us `VALUE_` which is incorrect. So I will update the json2jsii version with the fix you just approved and that should fix this. Until then we cannot merge this.

paulhcsun (approved):

test/import/fixtures/kube-prometheus.yaml
paulhcsun: lgtm, thanks for explaining. Added a do-not-merge tag, feel free to remove and merge once json2jsii version is updated

vinayak-kukreja: Great, thank you.

After the update it is,
```
/**
 * Match operation available with AlertManager >= v0.22.0 and takes precedence over Regex (deprecated) if non-empty.
 *
 * @schema AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType
 */
export enum AlertmanagerConfigSpecInhibitRulesSourceMatchMatchType {
  /** != */
  VALUE_NOT_EQUALS_TO = \\"!=\\",
  /** = */
  VALUE_EQUALS = \\"=\\",
  /** =~ */
  VALUE_EQUAL_TILDE = \\"=~\\",
  /** !~ */
  VALUE_NEGATION_TILDE = \\"!~\\",
}
```

PR #1928, opened 2024-02-08
fix: Add YAML document separator to each manifest

### LINKED ISSUE #2155: `cdk8s synth -p` fails to insert necessary document start markers
### Description of the bug:
If `cdk8s synth -p`  is used with any other `yamlOutputType` setting on the app than `YamlOutputType.FILE_PER_APP`, the YAML being output is missing some document start markers.

### Reproduction Steps:

1. Create a new `cdk8s` project
2. Adapt this project as per https://cdk8s.io/docs/latest/getting-started/#importing-constructs-for-the-kubernetes-api
3. Change `const app = new App();` to `const app = new App({ yamlOutputType: YamlOutputType.FILE_PER_RESOURCE });` in `main.ts`
4. Execute `npm run compile && cdk8s synth -p`
5. Observe the following output:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello-deployment-c8c7fda7
spec:
  replicas: 2
  selector:
    matchLabels:
      app: hello-k8s
  template:
    metadata:
      labels:
        app: hello-k8s
    spec:
      containers:
        - image: paulbouwer/hello-kubernetes:1.7
          name: hello-kubernetes
          ports:
            - containerPort: 8080
apiVersion: v1
kind: Service
metadata:
  name: hello-service-c8c17160
spec:
  ports:
    - port: 80
      targetPort: 8080
  selector:
    app: hello-k8s
  type: LoadBalancer
```

As you can see, this output is missing the YAML document start marker `---` before the second resource. 

### Other:

The reason is obviously that `synth -p` is not outputting documents one by one, but instead it runs `synth` into a temporary directory and then concatenated the files generated. The document start marker is missing at the exact spot where two files come together. For obvious reasons, this is not a problem with `YamlOutputType.FILE_PER_APP`. 

A simple solution could be adding a document start marker to each file or each resource by default (which is recommended anyway, e.g. in [yamllint](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.document_start)). 

---

This is :bug: Bug Report

rassie:
When output to STDOUT is requested, each manifest in an app is rendered separately and dumped to standard output. In case an app contains more than one chart, a document separator will be missing between adjacent charts, since there is no YAML document separator either at the beginning or at the end of an individual output. The last resource of the preceding chart and the first resource of the following chart will be merged into a random mess. By adding a YAML document separator after each manifest unconditionally, this can be cleanly avoided.

Fixes: cdk8s-team/cdk8s#2155


src/cli/cmds/synth.ts
```diff
@@ -94,6 +94,7 @@ class Command implements yargs.CommandModule {
         const app = await synthApp(command, tempDir, stdout, recordConstructMetadata);
         for (const f of app.manifests) {
           fs.createReadStream(f).pipe(process.stdout);
+          process.stdout.write('---\n');
```
iliapolo, line 97: This won't work as expected because the line just above it as asynchronous. It will print this before the manifests are printed. To do this we first need to await on the created pipeline:

```typescript
import { pipeline } from 'stream';
import { promisify } from 'util';

const asyncPipeline = promisify(pipeline);

await asyncPipeline(fs.createReadStream(f), process.stdout, { end: false });
process.stdout.write('---\n');
```

We should also have test to assert on the content of `process.stdout` using jest spies or something.

iliapolo (changes requested):

PR #2519, opened 2024-07-26
feat(import): csharp support

ScottKane:
Added support for calling `jsii-srcmak` with `csharp/dotnet` from `cdk8s import` (REGEX applied to convert names to PascalCase where possible)


iliapolo:
Looks like a we have another PR for this exact feature: https://github.com/cdk8s-team/cdk8s-cli/pull/2520.

@ScottKane How do you feel about reaching out to the author and deciding on the one PR we should be merging? :)

iliapolo (changes requested):
Missing integ test. See https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/test/integ/init.test.ts

iliapolo:
Ah ok this is just adding csharp to import and not as an init template. Still, we need a test for it then because its not covered by init tests.

iliapolo:
@ScottKane see https://github.com/cdk8s-team/cdk8s-cli/pull/2520#issuecomment-2341268022

ScottKane:
@iliapolo Integration test added and template added. I have also renamed everything to use the `csharp` convention

ScottKane:
also `jsii` seems to hard code target framework to `netcoreapp3.1` which isn't ideal but will solve that in the future

ScottKane:
@iliapolo is there anything else you want me to do here? afaik I addressed everything that was mentioned

iliapolo (approved):

PR #2520, opened 2024-07-26
feat(import): csharp support

haodeon:
Implements missing functionality for cdk8s-team/cdk8s#118

Adds `csharp` to srcmak options.

Have tested the generated code in a F# project, both importing `k8s@1.30.0` and some CRDs.


iliapolo:
Looks like a we have another PR for this exact feature: https://github.com/cdk8s-team/cdk8s-cli/pull/2519.

@haodeon How do you feel about reaching out to the author and deciding on the one PR we should be merging? :)

iliapolo (changes requested):
Missing integ test, see https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/test/integ/init.test.ts

iliapolo:
Ah ok this is just adding csharp to import and not as an init template. Still, we need a test for it then because its not covered by init tests.

haodeon:
I don't mind which PR gets merged.

My only feedback about the other PR is, where there are references to language it should be `csharp` as srcmak produces `csharp`.

ScottKane:
I am happy for either PR to get merged, I am currently using a fork in CI and would prefer a proper package to consume. If we can decide which we go with one of use can knock together a test and proceed.

@haodeon I went with the conventions of each code base, `jsii-srcmak` calls it `csharp` but elsewhere in cdk8s it's referred to as dotnet but I don't really care either way, it can all be consumed by any .NET language anyways.

iliapolo:
Ok so lets go with https://github.com/cdk8s-team/cdk8s-cli/pull/2519, for no good reason. As for the language, jsii refers to it as `dotnet` so lets go with that. Closing this one. Thanks @haodeon! and please provide any feedback you have on the other PR.

iliapolo:
@ScottKane ping me once you have test together or if you need guidance.

iliapolo:
> As for the language, jsii refers to it as dotnet so lets go with that.

Actually, CDK uses `charp`, so lets do that :)

PR #2728, opened 2024-09-14
fix(import)!: always generate classes with version suffixes

### LINKED ISSUE #1962: CRD versioning should be consistent. v1beta1, v1alpha1, v1 etc
### Description of the bug:

When using `cdk8s import`, if there are multiple versions of the CRD like `v1alpha1` & `v1beta1`, cdk8s uses v1alpha1 for main class name. Because of this, a lot of users are unknowingly using the v1alpha1 resource besides there is a newer version available.

### Reproduction Steps:

In this [example](https://github.com/external-secrets/external-secrets/blob/main/deploy/crds/bundle.yaml),

cdk8s generates 2 different versions for `ExternalSecret` CRD. However `v1alpha1` becomes the main class.

```ts
export class ExternalSecret extends ApiObject {
  /**
   * Returns the apiVersion and kind for "ExternalSecret"
   */
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'external-secrets.io/v1alpha1',
    kind: 'ExternalSecret',
  }
}

...

export class ExternalSecretV1Beta1 extends ApiObject {
  /**
   * Returns the apiVersion and kind for "ExternalSecretV1Beta1"
   */
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'external-secrets.io/v1beta1',
    kind: 'ExternalSecret',
  }
}
```

Or in this [example](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/standard), cdk8s generates 

```ts
export class HttpRoute extends ApiObject {
  /**
   * Returns the apiVersion and kind for "HTTPRoute"
   */
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'gateway.networking.k8s.io/v1',
    kind: 'HTTPRoute',
  }
}

...

export class HttpRouteV1Beta1 extends ApiObject {
  /**
   * Returns the apiVersion and kind for "HTTPRouteV1Beta1"
   */
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'gateway.networking.k8s.io/v1beta1',
    kind: 'HTTPRoute',
  }
}
```

In this case v1 class is the main class, much better.

There are other cases where it gets inconsistent. For example, when there are `v1` and `v2` CRDs, `v1` becomes the main class and `v2` gets prefix like `Somethingv2`. _However_ if there are `v2` and `v3` CRDs, `v2` CRD becomes main class with no prefix.

```ts
export class Host extends ApiObject {
  /**
   * Returns the apiVersion and kind for "Host"
   */
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'getambassador.io/v2',
    kind: 'Host',
  }
}
...

export class HostV3Alpha1 extends ApiObject {
  /**
   * Returns the apiVersion and kind for "HostV3Alpha1"
   */
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'getambassador.io/v3alpha1',
    kind: 'Host',
  }
}
```

### Environment:

  - **Framework Version**: `"cdk8s-cli": "2.198.105", "cdk8s": "^2.68.69",`
  - **OS**:

### Other:

Perhaps cdk8s can follow [Terraform k8s provider's approach](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/guides/versioned-resources) to version everything and remove unversioned class names.

---

This is :bug: Bug Report

shinebayar-g:
Fixes https://github.com/cdk8s-team/cdk8s/issues/1962

Tried to tackle https://github.com/cdk8s-team/cdk8s/issues/1962, turns out this is the only required change? Tested it on the following CRDs. Results seems correct.

```sh
npx cdk8s import --language typescript https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v1.1.0/config/crd/standard/gateway.networking.k8s.io_httproutes.yaml

npx cdk8s import --language typescript https://raw.githubusercontent.com/external-secrets/external-secrets/v0.10.3/deploy/crds/bundle.yaml
```

Unfortunately it'll be a breaking change.. :disappointed:

### DIFF
--- a/src/import/crd.ts
+++ b/src/import/crd.ts
@@ -98,11 +98,7 @@ export class CustomResourceDefinition {
     for (let i = 0; i < this.versions.length; i++) {
 
       const version = this.versions[i];
-
-      // to preseve backwards compatiblity, only append a suffix for
-      // the second version onwards.
-      const suffix = i === 0 ? '' : toPascalCase(version.name);
-
+      const suffix = toPascalCase(version.name);
       const types = new TypeGenerator({});
 
       generateConstruct(types, {
--- a/test/import/__snapshots__/import-crds-dev.test.ts.snap
+++ b/test/import/__snapshots__/import-crds-dev.test.ts.snap
@@ -85,19 +85,19 @@ Object {
     },
   },
   "types": Object {
-    "metapkgcrossplaneio.Configuration": Object {
+    "metapkgcrossplaneio.ConfigurationV1Alpha1": Object {
       "assembly": "metapkgcrossplaneio",
       "base": "cdk8s.ApiObject",
       "docs": Object {
         "custom": Object {
-          "schema": "Configuration",
+          "schema": "ConfigurationV1Alpha1",
         },
         "summary": "A Configuration is the description of a Crossplane Configuration package.",
       },
-      "fqn": "metapkgcrossplaneio.Configuration",
+      "fqn": "metapkgcrossplaneio.ConfigurationV1Alpha1",
       "initializer": Object {
         "docs": Object {
-          "summary": "Defines a \\"Configuration\\" API object.",
+          "summary": "Defines a \\"ConfigurationV1Alpha1\\" API object.",
         },
         "locationInModule": Object {
           "filename": "meta.pkg.crossplane.io.ts",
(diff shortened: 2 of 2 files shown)


iliapolo:
@shinebayar-g As this is a breaking change, we won't be able to merge it without releasing a new major version. 

Converting to draft until then.

PR #3161, opened 2025-01-21
chore: can skip tests during build by setting SKIP_TESTS=1

mrgrain:
This is only used by the package integrity check running outside the repo. While this check needs to replicate the release, it does not need to run tests.

To skip tests, set `SKIP_TESTS=1`

### DIFF
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -97,7 +97,7 @@ project.compileTask.spawn(schemas);
 
 // Allow skipping tests in build based on an env variable
 // This is only used by the package integrity check running outside the repo
-// While this check needs to replicate the release, it does not need to run tests 
+// While this check needs to replicate the release, it does not need to run tests
 project.testTask.addCondition("node -e \"if (process.env.SKIP_TESTS==='1') process.exit(1)\"");
 
 
--- a/.projenrc.ts
+++ b/.projenrc.ts
@@ -95,11 +95,10 @@ project.testTask.prependSpawn(installHelm);
 
 project.compileTask.spawn(schemas);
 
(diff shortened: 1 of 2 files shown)


.projenrc.ts
```diff
@@ -95,11 +95,10 @@ project.testTask.prependSpawn(installHelm);
 
 project.compileTask.spawn(schemas);
 
-// so that it works on windows as well
-// default projen uses $(npm pack) which fails
-project.packageTask.reset();
```
iliapolo, line 100: Nice! Thanks.

iliapolo (dismissed):

.projenrc.ts
```diff
@@ -95,11 +95,10 @@ project.testTask.prependSpawn(installHelm);
...
-project.packageTask.exec('mkdir -p dist/js');
-project.packageTask.exec('npm pack --pack-destination dist/js');
+// Allow skipping tests in build based on an env variable
+// This is only used by the package integrity check running outside the repo
+// While this check needs to replicate the release, it does not need to run tests
+project.testTask.addCondition("node -e \"if (process.env.SKIP_TESTS==='1') process.exit(1)\"");
```
iliapolo, line 101: Wait shouldn't exit with zero?

mrgrain: No that's right. It's just how projen conditions work:
- `Status==0` => success => run command
- `Status!=0` => failure => skip command
image

.projenrc.ts
mrgrain: Ah yes, that's not needed anymore apparently.

iliapolo (approved):

PR #3255, opened 2025-03-05
fix: Allowing helm oci versions with v prefix

clayrosenthal:
Allows specifying helm charts that version with `vX.Y.Z` rather than `X.Y.Z`. Helm supports this, but the import cli did not.


iliapolo (changes requested):
@clayrosenthal Thanks! 

Can you please add an example import url with `v` prefix to our list of tested URLs?

https://github.com/cdk8s-team/cdk8s-cli/blob/e1559ae7e134cf5ee1a06d1b9b45fd49e43c57be/test/import/import-helm.test.ts#L7-L14

PR #3458, opened 2025-06-09
feat(synth): option to specify chart name for helm format

MatthewCane:
Currently, when synthesising an app as a Helm chart the chart name is the base path of the chart.

This presents an issue when a cdk8s application is not in the root of the project, for example if the application is located in `/infrastructure/chart` within a larger project, the name of the chart will just be `chart`. This could be confusing.

This change adds an optional parameter `--chart-name` in the CLI or `chartName` in the config file that will allow users to override this value.

I have added tests which pass and tested locally and it works as expected.

This is my first PR for cdk8s so I am open to suggestions for improvements.


MatthewCane:
@iliapolo could I get some feedback on this please? Thanks!

test/synth/__snapshots__/synth-stdout.test.ts.snap
```diff
@@ -16104,6 +16104,142 @@ metadata:
 }
 `;
 
+exports[`Helm synthesis --chart-name is used when specified with all inputs from cli and no config file is present 1`] = `
+Object {
+  "Chart.yaml": "apiVersion: \\"v2\\"
+name: \\"cdk8s-synth-test\\"
```
iliapolo, line 16110: I believe this isn't the expected value here - it should be `custom-chart-name`? 

Its because you need to add `chartName` here:

https://github.com/cdk8s-team/cdk8s-cli/blob/9f54e63104543548b4d15b3e60fc8ff9a49f9cc8/test/synth/synth-stdout.test.ts#L1088

Otherwise, the cli input tests don't actually pass it to the synth handler.

iliapolo (changes requested):
Nice! Just one small fix to the test.

test/synth/__snapshots__/synth-stdout.test.ts.snap
MatthewCane: Good catch, thank you for the feedback! I have added that now and pushed the change.

test/synth/synth-stdout.test.ts
```diff
@@ -1009,6 +1009,119 @@ describe('Helm synthesis', () => {
...
+      },
+    ],
+    [
+      withOnlyConfigInputs,
+      {
+        config: {
+          synthConfig: {
+            format: SynthesisFormat.HELM,
+            chartVersion: '1.1.1',
+            chartName: 'custom-chart-name',
```
iliapolo: Dont think this belongs here - this test says its checking for: "default chart name is used when --chart-name **is not specified** %s".

So why add it? 

Note that my [previous comment](https://github.com/cdk8s-team/cdk8s-cli/pull/3458#discussion_r2176056853) was about adding `chartName` here:

https://github.com/cdk8s-team/cdk8s-cli/blob/ac7899abf1a293fd7d8dd4f43270474e3d8e496e/test/synth/synth-stdout.test.ts#L1088

So:

```ts
      chartApiVersion: options.chartApiVersion,
      chartVersion: options.chartVersion,
+    chartName: options.chartName
```

MatthewCane: It's been a while since I could look at this, does it look better now?

MatthewCane:
@iliapolo I have pushed an update, sorry for the delay, I've not been able to look at this for a while.

MatthewCane:
I am going to close this PR now as it hasn't been reviewed in several months. I think this is a really useful feature so if it ever gets attention again, feel free to reopen and merge it.

PR #3590, opened 2025-09-01
feat: init templates now require and use `uv` instead of `pipenv`

### LINKED ISSUE #2632: Support `uv`
### Description of the feature or enhancement:
Support using `uv` instead of `pipenv` when using the `cdk8s-cli`.

### Use Case:
I prefer not to use `pipenv`. I actually tend to use `pixi` which uses `uv`, but also adds support for `conda` packages, but I don't mind using `uv`. `pipenv` is slower and is legacy at this point AFAIK.

### Proposed Solution:
I could probably implement the change. It shouldn't be very big.

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

hajs:
This PR fixes cdk8s-team/cdk8s/issues/2632

pipenv was great when it was introduced but now nearly everyone moves on to uv: https://www.google.com/search?q=pipenv+vs+uv


iliapolo:
I'm a little hesitant merging this because its a behavior change...though I agree a desired one. 

Community, can we get some signals that this is a change you'd like merged?

Vad1mo:
the whole python ecosystem is moving to uv, I don't mind the change or backwards incompatibility (if there is any)

Vad1mo:
@hajs can you fix the DCO --> https://github.com/cdk8s-team/cdk8s-cli/pull/3590/checks?check_run_id=51432074438

templates/python-app/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "cdk8s-cli"
+version = "0.1.0"
+requires-python = ">=3.12"
```
iliapolo: @hajs is 3.12 really required? can we use 3.10? it still has a [year left](https://endoflife.date/python) to its security support.

hajs: Oh you are right. The cdk8s-package on pypi is is still compatible with 3.10. I just commited a change.

iliapolo:
@hajs the DCO is still failing. Can you please sign your commits? See instructions [here](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/CONTRIBUTING.md#developer-certificate-of-origin-dco).

iliapolo (approved):

PR #3731, opened 2025-12-08
feat: expose helm from Helm chart class

justin0u0:
The `Helm` class extends `Include` and expose `apiObjects`, which is useful for [patching](https://cdk8s.io/docs/latest/basics/escape-hatches/#target-programming-language-language) the synthesized output using JSON Patch.

Expose the `Helm` object so Helm chart class can be patched too.

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -57,4 +57,4 @@ jobs:
       fail-fast: false
       matrix:
         nodeVersion:
-          - 18
+          - 20
--- a/projenrc/integ.ts
+++ b/projenrc/integ.ts
@@ -33,7 +33,7 @@ export function addIntegTests(project: typescript.TypeScriptProject) {
     }
   }
 
-  // run all init tests on node 16
+  // run all init tests on all node LTS
   integWorkflow.addJob('init', {
     runsOn: ['${{ matrix.os }}'],
     strategy: {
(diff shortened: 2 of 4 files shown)


iliapolo (dismissed):

iliapolo:
@justin0u0 Looks like there are still some snapshots that needs updating.

justin0u0:
Hi @iliapolo, no problem.

I've rebased the branch and updated the snapshots. Also, I noticed that the jsii dependency recently started requiring Node.js 20 [^1]. Therefore, I also updated `projenrc.ts` to run `init-typescript-app` using Node.js 20; otherwise, it would fail.

[^1]: https://github.com/aws/jsii-compiler/pull/2507/changes#diff-dc3a048422937e4ece592ca654d2b32df0676e206b0bc7012e6b128795081bfa

.projenrc.ts
```diff
@@ -26,6 +26,7 @@ const project = new Cdk8sTeamTypeScriptProject({
   bin: {
     cdk8s: 'bin/cdk8s',
   },
+  minNodeVersion: '20.16.0',
```
iliapolo: Is this required as well for the integ tests to pass? Not sure I see why. 

This is harsh because it changes the `engines` directive of the package. We definitely want to update this soon but it requires a little bit more thought and I would prefer for it to be in a different PR.

justin0u0: Fixed. I noticed that `minNodeVersion` does not need to be updated.

Thanks!

iliapolo (approved):

PR #3798, opened 2026-02-03
fix(import): aggregate CRD imports from same API group

### LINKED ISSUE #3797: Multiple CRD imports from same API group overwrite each other
## Description

When importing multiple CRD files from the same Kubernetes API group (e.g., `compute.gcp.upbound.io`) using separate entries in `cdk8s.yaml` or multiple URLs on the command line, only the last CRD's types are generated. Earlier imports from the same API group are overwritten.

## Expected Behavior

All CRDs from the same API group should be consolidated into a single module, regardless of whether they come from separate files/URLs.

For example, when importing:
```yaml
imports:
  - https://example.com/compute.gcp.upbound.io_networks.yaml
  - https://example.com/compute.gcp.upbound.io_subnetworks.yaml
```

The resulting module `imports/io/upbound/gcp/compute/__init__.py` should contain both `Network` and `Subnetwork` classes.

## Actual Behavior

Only `Subnetwork` (from the second import) is available. The `Network` class from the first import is overwritten.

## Root Cause

In `src/import/dispatch.ts`, each import spec is processed independently via a loop:

```typescript
for (const importSpec of imports) {
  const importer = await matchImporter(importSpec, argv);
  await importer.import({ ... });  // Each one writes to disk, overwriting previous
}
```

While `ImportCustomResourceDefinition` correctly consolidates CRDs from the same API group within a single manifest file, there's no aggregation between separate import specs.

## Use Case

This is especially problematic when importing Crossplane provider CRDs, where each managed resource type is defined in its own CRD file (e.g., networks, subnetworks, clusters, node pools, etc.), but they all share the same API group.

## Workarounds

1. Manually download and combine CRD files before importing
2. Use module name prefix (`name:=url`) for each import (results in separate modules instead of one consolidated module)

## Proposed Solution

Aggregate all unprefixed CRD imports, download all manifests, combine them with YAML document separators (`---`), and process them together as a single `ImportCustomResourceDefinition`. This ensures CRDs from the same API group are properly consolidated.

CRD imports with a `moduleNamePrefix` would continue to be processed individually to preserve the user's intent for separate modules.

## Environment

- cdk8s-cli version: 2.x
- Affects all languages (TypeScript, Python, Java, Go)

mateo-moon:
## Summary

This PR fixes an issue where importing multiple CRD files from the same Kubernetes API group would cause earlier imports to be overwritten by later ones.

### The Problem

When importing multiple CRD files from the same API group (e.g., `compute.gcp.upbound.io`) using separate entries:

```yaml
imports:
  - https://example.com/compute.gcp.upbound.io_networks.yaml
  - https://example.com/compute.gcp.upbound.io_subnetworks.yaml
```

Only `Subnetwork` (from the second import) would be available in the generated module. The `Network` class from the first import was overwritten.

### The Solution

- Aggregate all unprefixed CRD imports before processing
- Download all manifests and combine them with YAML document separators (`---`)
- Process combined manifest as a single `ImportCustomResourceDefinition`
- CRD imports with `moduleNamePrefix` are still processed individually (preserves user intent)

### Changes

- **`src/import/dispatch.ts`**: Added `categorizeImports()` function to group imports by type, then process unprefixed CRDs together
- **`src/import/crd.ts`**: Added `fromSpecs()` and `fromManifest()` static methods
- **`test/import/import-crd.test.ts`**: Added 3 test cases for the aggregation behavior
- **`test/import/fixtures/`**: Added test fixture files for same-group CRD testing

### Use Case

This is especially important when importing Crossplane provider CRDs, where each managed resource type is in its own CRD file but shares the same API group.

Fixes #3797

---

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*


hpedrorodrigues:
Hey 👋🏻 
How can I help move this forward? I'm also interested in these changes.

mateo-moon:
@hpedrorodrigues thanks for the offer to help, and apologies for the slow reply. I have just rebased this branch onto the current 2.x head; the import tests and the full local test suite pass. This still addresses #3797, which remains open. @iliapolo could you approve the workflow run and take a look when you get a chance?

PR #3801, opened 2026-02-03
fix: resolve TypeScript built-in type name collisions in CRD imports

### LINKED ISSUE #3800: CRD import fails when kind name matches TypeScript built-in type (e.g., Record)
## Description

When importing a CRD whose `kind` name matches a TypeScript built-in utility type (e.g., `Record`, `Map`, `Set`), the generated TypeScript code fails to compile due to naming conflicts.

## Steps to Reproduce

1. Import a CRD with kind `Record` (e.g., Cloudflare DNS Record CRD):

```yaml
# cdk8s.yaml
imports:
  - https://raw.githubusercontent.com/cdloh/provider-cloudflare/main/package/crds/dns.cloudflare.upbound.io_records.yaml
```

2. Run `cdk8s import`

3. Try to compile the generated TypeScript

## Expected Behavior

The generated code should compile without errors.

## Actual Behavior

TypeScript compilation fails with errors like:

```
error TS2315: Type 'Record' is not generic.
```

This happens because:
1. The generated code creates a class named `Record` (from the CRD kind)
2. The `toJson_*` functions use `Record<string, any>` as return types
3. The class `Record` shadows TypeScript's built-in `Record<K, V>` utility type

## Affected CRDs

Any CRD with a `kind` that matches a TypeScript built-in type:
- `Record` (e.g., Cloudflare DNS Record, various other providers)
- Potentially `Map`, `Set`, `Array`, `Object`, `String`, `Number`, `Boolean`, `Function`, `Promise`, etc.

## Environment

- cdk8s-cli version: 2.x
- Node.js version: 20+
- TypeScript version: 5.x

mateo-moon:
## Summary

Fixes #3800

When a CRD's kind name matches a TypeScript built-in utility type (e.g., `Record`, `Map`, `Set`), the generated code creates a class that shadows the built-in type. This causes TypeScript compilation errors when the generated `toJson_*` functions use types like `Record<string, any>`.

## Changes

Added a post-processing step in `src/import/crd.ts` that:

1. Maintains a list of TypeScript built-in utility types that may conflict
2. Detects when a CRD kind matches one of these types
3. Adds a type alias (e.g., `type JsonRecord<K extends keyof any, T> = { [P in K]: T };`) at the beginning of the generated file
4. Replaces usages of the built-in type with the alias (e.g., `Record<string, any>` → `JsonRecord<string, any>`)

## Example

For Cloudflare's DNS `Record` CRD, the generated code now includes:

```typescript
// Type alias to avoid collision with the Record class defined below
type JsonRecord<K extends keyof any, T> = { [P in K]: T };

export class Record extends ApiObject {
  // ...
}

export function toJson_RecordProps(obj: RecordProps | undefined): JsonRecord<string, any> | undefined {
  // ...
}
```

## Test Plan

- [ ] Verified TypeScript compilation passes
- [ ] Manual testing with Cloudflare DNS Record CRD
- [ ] Consider adding unit tests for the `fixBuiltinTypeConflicts` function

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


mateo-moon:
Rebased onto the current 2.x head. While running this patch in production we found three follow-up issues, and the fixes are now included here: the type alias is emitted only once for multi-version CRDs, runtime calls such as Object.entries/keys/values are redirected to a preserved _Object alias, and those runtime replacements are applied to every CRD generated into the same file. The full local test suite passes. This still addresses #3800, which remains open. @iliapolo could you approve the workflow run and review?

PR #3833, opened 2026-03-17
feat(cli): bump default `cdk8s-plus` dependency to 33

iliapolo:
Version 28 is EOL long ago, 33 is the lowest version still supported. 

https://endoflife.date/kubernetes

This also resolves a transitive mnimatch vulnerability.

### DIFF
--- a/projenrc/integ.ts
+++ b/projenrc/integ.ts
@@ -102,7 +102,7 @@ function runSteps(tasks: string[], nodeVersion: string, python: boolean, go: boo
       name: 'Set up Go',
       uses: 'actions/setup-go@v6',
       with: {
-        'go-version': '1.24',
+        'go-version': '1.25',
       },
     });
   }
--- a/templates/typescript-app/package.json
+++ b/templates/typescript-app/package.json
@@ -22,11 +22,11 @@
   },
   "devDependencies": {
     "cdk8s-cli": "{{ cdk8s_cli_spec }}",
-    "@types/node": "^14",
-    "@types/jest": "^27",
-    "jest": "^27",
-    "ts-jest": "^27",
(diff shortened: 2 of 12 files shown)


projenrc/integ.ts
```diff
@@ -102,7 +102,7 @@ function runSteps(tasks: string[], nodeVersion: string, python: boolean, go: boo
       name: 'Set up Go',
       uses: 'actions/setup-go@v6',
       with: {
-        'go-version': '1.24',
```
iliapolo, line 105: `jsii-pacmak` now [requires 1.25](https://github.com/aws/jsii/pull/4895/changes/77d81b6075d570d9f1c1aa4596f68b61580adad4).

templates/typescript-app/package.json
```diff
@@ -17,16 +17,16 @@
...
+    "cdk8s-plus-33": "^{{ cdk8s_plus_version }}",
     "constructs": "^{{ constructs_version }}"
   },
   "devDependencies": {
     "cdk8s-cli": "{{ cdk8s_cli_spec }}",
-    "@types/node": "^14",
-    "@types/jest": "^27",
-    "jest": "^27",
-    "ts-jest": "^27",
-    "typescript": "^4.9.5",
```
iliapolo, line 29: These are actually unrelated to our change and cause integ tests to currently fail on every PR. 

> See https://github.com/cdk8s-team/cdk8s-cli/actions/runs/23186126253/job/67369959477?pr=3832

It happens because the init templates (intentionally) install latest versions of our dependencies - which introduced new limitations we weren't aware of due to the lack of canaries.

Just bumping here to use the latest of everything.

otaviomacedo (approved):

